PR #25, opened 2025-12-11
floating player now loads on app opening with the correct surah and p…

OKKHALIL3:
Floating player now saves state of play even after closing the app fully.


osmansaeday (approved):
LGTM

PR #190, opened 2026-04-09
feat: integrate Bayaan backend API — replace bundled data with live API

OKKHALIL3:
## Summary

- **Live API integration**: `dataService.ts` now fetches all reciter data from the Bayaan backend API instead of the bundled `reciters.json`. Uses a cache-first strategy — returns `AsyncStorage` cache immediately on subsequent launches, refreshes from API in the background.
- **Removed bundled data**: `data/reciters.json` deleted from the app. The backend is now the single source of truth for all reciter and rewayat metadata.
- **Mutable RECITERS array**: `data/reciterData.ts` exports an empty mutable array. `dataService` populates it in-place after fetching so all 20+ files that import `RECITERS` automatically see live data without any import changes.
- **Community fork support**: If `EXPO_PUBLIC_BAYAAN_API_KEY` is not set, the app skips API calls and falls back to the empty array. Contributors can obtain a community key from `CONTRIBUTING.md`.
- **reciterStore crash fix**: `getDefaultReciter()` now returns a safe empty placeholder when `RECITERS` is empty (first launch before API data loads). Added `refreshDefaultReciter()` action — called by `dataService` after populating `RECITERS` so the store picks up the real default reciter (Mishary Alafasi) without a restart.
- **Env vars via Expo**: Switched from `react-native-dotenv` (`@env`) to Expo's native `EXPO_PUBLIC_` convention (`process.env`). The dotenv babel plugin was interfering with `react-native-reanimated`'s workletization, causing a fatal crash on launch.

## Env vars required in `Bayaan/.env`

```
EXPO_PUBLIC_BAYAAN_API_URL=https://bayaan-backend-production.up.railway.app
EXPO_PUBLIC_BAYAAN_API_KEY=<raw key from /admin/keys>
```

## Files changed

| File | Change |
|------|--------|
| `services/dataService.ts` | Full API integration with cache-first strategy |
| `data/reciterData.ts` | Removed JSON import, RECITERS is now empty mutable array |
| `data/reciters.json` | Deleted — moved to `Backend--bayaan/data/` |
| `store/reciterStore.ts` | Crash guard + refreshDefaultReciter action |
| `babel.config.js` | Removed react-native-dotenv plugin |
| `.env.example` | Updated to EXPO_PUBLIC_ var names |
| `CONTRIBUTING.md` | Added API key setup instructions for contributors |

## Test plan

- [ ] Fresh install (no cache): app fetches from API on launch, reciters load
- [ ] Second launch: reciters load instantly from AsyncStorage cache
- [ ] No `EXPO_PUBLIC_BAYAAN_API_KEY` set: app starts without crashing, shows empty state
- [ ] Audio playback works for mp3quran and R2 CDN reciters
- [ ] Default reciter (Mishary Alafasi) is set correctly after data loads


osmansaeday:
## PR #190 Review: Integrate Bayaan Backend API for Reciter Data

**4 commits** | +253 / -34,323 lines | Deletes bundled `reciters.json`, fetches reciter data from backend API

### Overview

The migration from a 34K-line bundled JSON file to a backend API with AsyncStorage caching and background refresh is a solid architectural direction. The shared mutable array populated via `splice` is a pragmatic approach given the 44+ files that import `RECITERS`. However, there are several bugs and inconsistencies that should be addressed before merging.

---

### Bugs

**1. `react-native-dotenv` added instead of removed**
The final commit message says "react-native-dotenv removed from devDependencies," but the `package.json` diff shows it being **added**. The babel config has no `module:react-native-dotenv` plugin, so this package is unused dead weight. It should be removed from `devDependencies`.

**2. `Authorization: Bearer undefined` sent when no API key is set**
In `dataService.ts`, the `Authorization` header is always included:
```ts
headers: {
  Authorization: `Bearer ${API_KEY}`,
```
When `EXPO_PUBLIC_BAYAAN_API_KEY` is unset, this sends `Authorization: Bearer undefined`. This will likely cause a 401 rather than the graceful "no-key" fallback described in CONTRIBUTING.md. The header should be conditionally included only when `API_KEY` is truthy.

**3. CONTRIBUTING.md references wrong env var names**
CONTRIBUTING.md says:
```
BAYAAN_API_URL=https://api.bayaan.app
BAYAAN_API_KEY=<community key>
```
But `.env.example` uses `EXPO_PUBLIC_BAYAAN_API_URL` and `EXPO_PUBLIC_BAYAAN_API_KEY`. Contributors following CONTRIBUTING.md will set the wrong vars and silently fall through to no data.

**4. CONTRIBUTING.md claims fallback to bundled `reciters.json` — but that file is deleted**
> "the app falls back to the bundled `data/reciters.json` automatically"

This is false after this PR. `reciters.json` is deleted. If the API fails and there's no cache, the app returns an empty array and users see zero reciters. The docs need to reflect this new reality.

**5. `types/env.d.ts` adds dead `@env` declarations**
The PR adds `BAYAAN_API_URL` and `BAYAAN_API_KEY` to the `@env` module declaration, but the code now imports via `process.env.EXPO_PUBLIC_*`. These declarations are unused and misleading.

---

### Design Concerns

**6. Background refresh only compares array length**
`refreshRecitersInBackground` skips the update if `apiReciters.length === cachedCount`:
```ts
if (apiReciters.length > 0 && apiReciters.length !== cachedCount) {
```
If a reciter's server URL, name, or rewayat data changes without the total count changing, the update is silently dropped. Consider comparing a hash or version field from the API, or just always writing the fresh data.

**7. First-launch experience with no key and no network**
On first launch with no API key (or network failure), `getAllReciters()` returns `[]`. The `reciterStore` returns a placeholder reciter with `id: ''`. The 44 files that depend on `RECITERS` being populated will all see an empty array. This may cascade into broken UI states (empty search, no default reciter, etc.). Consider:
- Showing an error/retry prompt to the user
- Bundling a minimal fallback dataset (even just the top 10 reciters)

**8. Hardcoded Railway fallback URL**
```ts
const API_BASE = BAYAAN_API_URL ?? 'https://bayaan-backend-production.up.railway.app';
```
If a community contributor omits the env var, they silently hit the production Railway URL. The CONTRIBUTING.md says the URL is `https://api.bayaan.app`. These should be consistent — or the fallback should be removed so missing config fails explicitly.

---

### Minor / Nits

- The `useReciterStore` import in `dataService.ts` creates a circular-ish dependency path (`dataService → reciterStore → reciterData`). It works because of JS module caching, but worth noting.
- Removed comments in `filterReciters` ("Filter by style", "Filter by surah", etc.) were actually useful for readability — their removal is cosmetic churn.
- The paginated fetch (`PAGE_SIZE = 200`) doesn't handle partial failures — if page 2 of 3 fails, all data from page 1 is lost.

---

### Summary

The core architecture is sound — cache-first with background refresh and in-place mutation of a shared array is a practical approach. But the PR has **documentation/code mismatches** (wrong env var names, false fallback claims, contradictory commit message about dotenv) and a **bug with the auth header** that will break the "no key" fallback path. I'd recommend addressing items 1–5 before merging.

---
_Generated by [Claude Code](https://claude.ai/code)_

OKKHALIL3:
All review items have been addressed in b6ec8ff:

**Bugs fixed:**
- `react-native-dotenv` removed from `devDependencies` — it was never used after switching to `EXPO_PUBLIC_`
- `Authorization: Bearer undefined` fixed — header is now conditionally included only when `API_KEY` is truthy; `getAllReciters()` returns early with a warning if either `API_BASE` or `API_KEY` is missing
- Hardcoded Railway fallback URL removed from `API_BASE` — missing config now fails explicitly instead of silently hitting production
- Background refresh now always writes fresh API data regardless of count changes

**Docs fixed:**
- `CONTRIBUTING.md` — env var names corrected to `EXPO_PUBLIC_BAYAAN_API_URL` / `EXPO_PUBLIC_BAYAAN_API_KEY`
- `CONTRIBUTING.md` — removed false claim about falling back to bundled `reciters.json`; no key now correctly documented as empty state
- `types/env.d.ts` — removed dead `BAYAAN_API_URL` and `BAYAAN_API_KEY` declarations from the `@env` module

**Re: the API serving audio directly** — correct, the backend only resolves metadata and returns CDN/mp3quran server URLs. Audio bytes stream directly from `cdn.thebayaan.com` or `mp3quran.net` — the backend is never in that path.

osmansaeday (approved):
lgtm

OKKHALIL3:
## Shareable deep links added to this branch

Merged `feat/deep-links` in — this PR now also includes shareable HTTPS deep links via `https://api.thebayaan.com/share/*` (backend already deployed).

**What's included:**
- `app/share/**` — Expo Router receiver screens for reciter, surah, mushaf, adhkar, playlist
- `app.config.js` — `associatedDomains` for iOS Universal Links + Android App Links intent filters
- `utils/shareUtils.ts` — URL builders and cross-platform `Share.share()` wrapper
- Share button added to reciter profile header (both iOS glass and Android paths)
- `data/reciterData.ts` — added `slug?` field to `Reciter` interface

**To test before merging:**
- Paste `https://api.thebayaan.com/share/reciter/al-husary` in WhatsApp → should show OG preview card
- Install a dev build on physical device → tap a share link → should open the correct screen directly
- Tap share icon on any reciter profile → share sheet should appear with the correct URL

PR #233, opened 2026-05-02
Feature/qcf v2 mushaf

OKKHALIL3:


OKKHALIL3:
@osmansaeday Based on the new test flight push, I think this is ready to go. What do you think?

PR #235, opened 2026-05-04
docs(rfc): RFC-007 — app-layer multi-tenancy seams (branding, featureFlags, CatalogProvider)

omar-zarka:
## Summary

Proposes RFC-007: three thin config seams so a sibling fork can override Bayaan's app identity, feature set, and reciter catalog without touching application code.

- `config/branding.js` (+ `.d.ts`) — app name, URLs, bundle IDs, scheme. CJS so `app.config.js` can `require()` it.
- `config/featureFlags.ts` — `isFeatureEnabled('flagName')`. All flags default `true`.
- `types/CatalogProvider.ts` + `data/defaultCatalogProvider.ts` — abstraction over the static `RECITERS` array.

**Zero behavior change for Bayaan.** The defaults are Bayaan's current hardcoded values; no app code imports from these files yet. Call-site migrations (`utils/shareUtils.ts`, settings screen, `services/emailService.ts`) are deliberate follow-on goodwill PRs per the RFC migration table.

## Context

Qariah is a sibling app forking Bayaan to ship a curated female-reciter catalog. Today every divergence is a direct file edit that conflicts on the weekly upstream merge. RFC-007 is the seam layer that makes the fork sustainable — and the same seams are useful to anyone else forking Bayaan.

This is the first cross-fork RFC (separate track from the platform-package extraction RFCs 001–006). It does not block or require any of those.

Happy to discuss scope, alternatives, or break this into smaller PRs (e.g., branding-only first, featureFlags second, CatalogProvider third) if that's easier to land.

## Test plan

- [ ] `npx tsc --noEmit` is green from repo root (verified locally)
- [ ] iOS build produces identical artifact to develop baseline
- [ ] Android build produces identical artifact to develop baseline
- [ ] No runtime behavior change (verified by no app code importing from new files)

### DIFF
--- a/config/branding.d.ts
+++ b/config/branding.d.ts
@@ -1,0 +1,23 @@
+/** Catalog source config for the active branding. */
+export interface BrandingCatalogConfig {
+  source: 'bundled' | 'remote';
+  /** Path to bundled JSON catalog file; undefined = use data/reciters-fallback.json. */
+  fallbackPath?: string;
+}
+
+/** App identity values that vary across forks. */
+export interface Branding {
+  appName: string;
+  appSlug: string;
+  urlScheme: string;
+  bundleId: {ios: string; android: string};
+  supportUrl: string;
+  termsUrl: string;
+  privacyUrl: string;
+  shareBaseUrl: string;
+  emailProductName: string;
+  catalog: BrandingCatalogConfig;
+}
+
+declare const branding: Branding;
+export default branding;
--- a/config/branding.js
+++ b/config/branding.js
@@ -1,0 +1,26 @@
+/**
+ * Bayaan app identity config.
+ *
+ * CJS (not ESM) — app.config.js calls require() before Expo CLI transpiles.
+ * Forks override by replacing this file; the shape is typed in branding.d.ts.
+ *
+ * See docs/rfcs/007-app-layer-multi-tenancy.md for the design rationale.
+ */
+module.exports = {
+  appName: 'Bayaan',
+  appSlug: 'Bayaan',
+  urlScheme: 'bayaan',
+  bundleId: {
+    ios: 'com.bayaan.app',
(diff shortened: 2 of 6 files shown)


OKKHALIL3 (changes requested):
Good foundation for the seam layer — the branding config, feature flags, and CI
  workflow are all clean and the zero-behavior-change guarantee for Bayaan is the right
  call for a first pass.

  One thing to fix before merging: types/CatalogProvider.ts has a JSDoc comment saying
  forks can call setCatalogProvider() at app boot, but that function doesn't exist
  anywhere. Qariah will hit this the moment they try to inject their own reciter catalog.
   Either add the implementation (module-level provider variable + setter, ~10 lines in
  data/defaultCatalogProvider.ts) or update the doc to say override is file-replacement
  rather than runtime injection.

  Everything else looks good to go.

omar-zarka:
Thanks for the catch, @OKKHALIL3 — good spot.

Added `setCatalogProvider()` and `getCatalogProvider()` to `data/defaultCatalogProvider.ts` in the latest commit (`c665e1d`):

- Module-level `_activeProvider` variable, initialized to `defaultCatalogProvider` (so Bayaan's behavior is unchanged — it never calls `setCatalogProvider()`).
- `setCatalogProvider(provider)` — intended to be called once at app boot before any component reads the catalog.
- `getCatalogProvider()` — what app code should call instead of importing `defaultCatalogProvider` directly; returns whichever provider is active.

Updated the JSDoc in `types/CatalogProvider.ts` to reference both exports precisely.

`tsc --noEmit` is clean. Let me know if you'd prefer a different shape (e.g. a class, or exporting a `catalogProvider` re-export that's reassignable) — happy to adjust.

OKKHALIL3 (dismissed):
JazakAllahu Khairun for the contribution.

PR #236, opened 2026-05-05
docs(rfc): RFC-004 — audio seam (PlayerController port + adapter scaffold)

omar-zarka:
## What this PR is

The fourth in the RFC series and the first to introduce a feature package. Lands ADR-0001's audio seam — five port interfaces plus a stub adapter — as a new `@bayaan/audio` workspace package. **No app code changes.** Services in `services/audio/*` are untouched; this PR sits alongside them.

The actual refactor of `AudioCoordinator`, `ExpoAudioProvider`, and `LockScreenService` to consume these ports is RFC-005, deliberately split off so this PR is reviewable in isolation.

Full RFC: [`docs/rfcs/004-audio-seam.md`](docs/rfcs/004-audio-seam.md). ADR-0001 (the design rationale) lives in the planning repo at [`adr/0001-audio-extraction-seam.md`](https://github.com/omar-zarka/bayaan-platform/blob/main/adr/0001-audio-extraction-seam.md).

### `packages/bayaan-audio/` — new workspace package

- `src/ports.ts` — five port interfaces:
  - `PlayerController<TTrack>` — state / progress / persistence bridge (replaces today's bidirectional `ExpoAudioProvider ↔ playerStore` coupling)
  - `CoordinatorHooks` — replaces the lazy-`require()` smell in `AudioCoordinator`
  - `PlayerEventSink<TTrack>` — analytics hook, with `NullPlayerEventSink` no-op default
  - `TimestampProvider` — ayah-timing fetch
  - `LockScreenMetadataSource` — subscription source for OS lock-screen metadata
  - `BayaanAudioConfig<TTrack>` — the consumer config the future provider takes at init
- `src/adapters/expo-audio-player-controller.ts` — stub `PlayerController` class (no-op bodies; every method has a `// RFC-005: ...` comment naming the migration that fills it in)
- `src/__tests__/ports.test.ts` — 18 tests covering port shape, the `NullPlayerEventSink` no-op, and realistic call ordering through the stub adapter

### `packages/bayaan-test-utils/` — additive

- `src/audio/mock-audio-player.ts` — `MockAudioPlayer` fixture mirroring `expo-audio`'s `AudioPlayer` surface plus deterministic `simulate*` drivers. Used by the adapter tests today; reusable in RFC-005's service-layer tests.
- 6 new tests for the fixture itself.
- `src/index.ts` re-exports `MockAudioPlayer` and its types.

### `package-lock.json`

Workspace symlinks registered at `node_modules/@bayaan/audio -> packages/bayaan-audio` and updated for the test-utils additions.

## What it does not do

- **No `services/audio/*` changes.** Provider, coordinator, services all untouched.
- **No `playerStore` decoupling.** Bidirectional coupling stays; RFC-005 fixes it.
- **No `expo-audio` import in the package.** `services/audio/*` keeps owning the engine until RFC-005 PR 6 (the move).
- **No characterization tests for current audio behavior.** ADR-0001's PR 1 covers those — they need real-device fidelity for some flows (lock-screen metadata sync, AppState transitions, interruption rate-reset) and are best authored against the running app, not a new package. They land alongside RFC-005's first refactor PR.
- **No Qariah adoption.** Qariah's `PlayerController` implementation lives in the Qariah repo and lands when RFC-005 ships.

## Verification

| Check | Result |
| --- | --- |
| `npm install` | Workspace symlink created at `node_modules/@bayaan/audio -> packages/bayaan-audio` |
| `npx jest packages/bayaan-audio packages/bayaan-test-utils` | **33/33 pass in ~0.9s** |
| `npm run test:ci` (full suite) | **8 suites, 77 tests pass in ~1.5s, 0 failures** (53 pre-existing + 18 new ports + 6 new MockAudioPlayer) |
| `npx tsc --noEmit` | **Zero new errors** vs `develop` baseline |
| Local iOS + Android builds | _Verification comment to follow with logs and timing._ |

## One bounded decision called out for review

**Where the stub adapter lives.** ADR-0001 says "package from day one" so the dependency direction is established before any consumer adopts. This PR follows that: the stub lives at `packages/bayaan-audio/src/adapters/expo-audio-player-controller.ts`. Two alternatives if you'd prefer either:

- **(a)** Keep it in-package — what this PR currently does.
- **(b)** Stage in `services/` until RFC-005 fills it in. Trade-off: dependency direction stays unenforced for one extra RFC cycle, but the migration story is "in-place refactor" rather than "match the new shape".
- **(c)** Drop the stub entirely and ship ports-only. Trade-off: package no longer has anything class-based to test against; contract tests would be type-only.

I think (a) is right but happy to redirect.

## Why this is its own PR

Three reasons to land the seam before any refactor that consumes it:

1. **Surface review can happen now.** The five interfaces are the contract that drives every subsequent PR in the audio extraction series. Catching shape problems at the type-only stage is cheap; catching them mid-refactor is expensive.
2. **The diff is bounded and reviewable in isolation.** If a later RFC is rejected, this package keeps compiling and harms nothing.
3. **Tidy First.** Mixing the seam definition with the first refactor would bundle "what shape" with "how to migrate to it" in one diff. Each PR should answer one question.

## Risk if this lands and RFC-005 is rejected

The package contains no app-side imports. If RFC-005 never ships, the only consequence is a small workspace package with port interfaces nobody uses, sitting alongside `@bayaan/types` and `@bayaan/test-utils`. No code outside the package depends on it after this PR.

## Open questions called out in the RFC

- **Adapter location** (a/b/c above).
- **Should `BayaanAudioConfig.eventSink` be required-explicit or optional-with-default?** This RFC ships it as required (forces the consumer to acknowledge the analytics decision; pass `NullPlayerEventSink` to opt out). Easy to flip.
- **Schema version field on `BayaanAudioConfig`?** Probably premature — flagging only.

---

Happy to iterate on the port shapes, the stub adapter location, or the required-vs-optional `eventSink` decision.

### DIFF
--- a/docs/rfcs/004-audio-seam.md
+++ b/docs/rfcs/004-audio-seam.md
@@ -1,0 +1,141 @@
+# RFC-004 — Audio seam: `PlayerController` port + adapter scaffold
+
+| Field | Value |
+| --- | --- |
+| Status | Proposed |
+| Author | Omar Zarka (Qariah) |
+| Date | 2026-05-05 |
+| Depends on | [RFC-001](./001-workspace-and-rfc-convention.md), [RFC-002](./002-bayaan-types-package.md), [RFC-003](./003-ci-gates-and-test-utils.md) |
+| Drives | RFC-005 (audio services refactor + extraction) |
+
+## TL;DR
+
+Ship the `@bayaan/audio` workspace package with **ports only** — the five
+interfaces from ADR-0001 plus a stub `ExpoAudioPlayerControllerAdapter`.
+**No app code changes.** No imports from this package by `services/audio/*`.
+The diff is exclusively under `packages/bayaan-audio/` and a small addition
+to `@bayaan/test-utils` (`MockAudioPlayer` fixture).
+
+This is ADR-0001's PR 2 ("port interfaces + no-op defaults"). The actual
+refactor of `AudioCoordinator`, `ExpoAudioProvider`, and `LockScreenService`
+to consume these ports is RFC-005, deliberately split off so this PR stays
+mergeable on a single bounded review.
+
+## What this PR adds
+
+### `packages/bayaan-audio/`
+
+- `src/ports.ts` — the five interfaces from [ADR-0001](https://github.com/omar-zarka/bayaan-platform/blob/main/adr/0001-audio-extraction-seam.md):
+  - `PlayerController<TTrack>` — state / progress / persistence bridge
+  - `CoordinatorHooks` — replaces the lazy-`require()` smell in `AudioCoordinator`
+  - `PlayerEventSink<TTrack>` — analytics hook (with `NullPlayerEventSink` no-op)
+  - `TimestampProvider` — ayah-timing fetch
+  - `LockScreenMetadataSource` — subscription source for OS lock-screen metadata
+  - `BayaanAudioConfig<TTrack>` — the full consumer config the future provider takes at init
+- `src/adapters/expo-audio-player-controller.ts` — stub class implementing
+  `PlayerController` end-to-end (no-op bodies). Demonstrates the seam compiles
+  and is implementable as a class. Real wiring lands in RFC-005.
+- `src/__tests__/ports.test.ts` — port shape contract: every method exists,
+  every state is accepted without throwing, the `NullPlayerEventSink` is a
+  full no-op, the stub adapter survives the realistic call ordering during
+  playback.
+- `package.json`, `tsconfig.json`, `jest.config.js`, `README.md` — same
+  shape as `@bayaan/types` and `@bayaan/test-utils`. No build step;
+  `main` and `types` both point at `src/index.ts`.
+
+### `packages/bayaan-test-utils/`
+
+- `src/audio/mock-audio-player.ts` — `MockAudioPlayer` test fixture mirroring
+  `expo-audio`'s `AudioPlayer` surface (play / pause / seekTo / replace /
+  setRate / setVolume / addListener) plus deterministic `simulate*` drivers.
+- `src/__tests__/mock-audio-player.test.ts` — 6 tests covering initial state,
+  listener subscription / removal, the natural-end (`didJustFinish`) transition,
+  source replacement, and jest-mock surface.
+- `src/index.ts` — re-export `MockAudioPlayer` and its types.
+
+## What this PR does not do
+
+- **No `services/audio/*` changes.** Provider, coordinator, services all untouched.
+- **No `playerStore` decoupling.** Bidirectional coupling stays; RFC-005 fixes it.
+- **No `expo-audio` import in the package.** `services/audio/*` keeps owning
+  the engine until RFC-005 PR 6 (the move).
+- **No characterization tests for current audio behavior.** ADR-0001's PR 1
+  covers those — they need real-device fidelity for some flows (lock-screen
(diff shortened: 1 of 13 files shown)


omar-zarka:
## Build verification — iOS confirmed, Android skipped (rationale below)

### iOS — \`npm run ios\` ✅

Built and ran locally on iPhone 16 Pro / iOS 26.4 sim (M-series Mac, Xcode 26.4.1).

- \`xcodebuild\`: **Build Succeeded** (0 errors, 0 warnings in build output)
- App installed under \`com.bayaan.app\` and opened on iPhone 16 Pro
- **Playback smoke-tested manually:** played one recitation end-to-end; audio works as on \`develop\`
- New workspace symlink at \`node_modules/@bayaan/audio -> packages/bayaan-audio\` resolves cleanly through Metro (the package is consumed only by package-internal tests today; no app-side imports)

Same Pods sandbox sync gotcha as RFC-002 / RFC-003 after a branch switch — \`pod install\` first, then \`expo run:ios\` succeeded. \`ios/Podfile.lock\` is untouched in this PR (the regenerated lock was reverted before commit; deltas were just local CocoaPods checksum churn).

### Tests + types

| Check | Result |
| --- | --- |
| \`npx jest packages/bayaan-audio packages/bayaan-test-utils\` | **33/33 pass in ~0.9s** |
| \`npm run test:ci\` (full suite) | **8 suites, 77 tests pass in ~1.5s, 0 failures** (53 pre-existing + 18 new ports + 6 new MockAudioPlayer) |
| \`npx tsc --noEmit\` | **Zero new errors** vs \`develop\` baseline |
| \`npx prettier --check\` on new files | clean |

### Android — \`npm run android\` ⚠️ skipped

Did not run \`expo run:android\` for this PR. Neither a physical device nor a running emulator was attached at verification time, and the cost of booting + cold-building Android (~10–15 min) didn't seem proportional to what the build would actually verify here.

**Why I think skipping is acceptable for this specific RFC:**

- \`@bayaan/audio\` is a **TypeScript-only** workspace package today. No \`expo-audio\` import, no native module, no Gradle plugin, no Android/iOS-specific code path. Every test runs in jest under \`jest-expo\` preset.
- **No app-side import** of \`@bayaan/audio\` exists in this PR. \`services/audio/*\` is byte-for-byte unchanged. The native build doesn't see the new package at all — it would only verify that adding a workspace symlink doesn't break Gradle's autolinking, which is the same thing RFC-002 and RFC-003 verified successfully and which has not regressed since.
- The \`MockAudioPlayer\` addition to \`@bayaan/test-utils\` is also pure TypeScript; it imports nothing native.

**Where Android verification becomes load-bearing:** RFC-005's first refactor PR (the one that wires \`AudioCoordinator\` to \`CoordinatorHooks\` and removes the lazy \`require()\`) is the first time this series touches code the native build cares about. Android local + smoke test of audio playback should be required there, and I'll plan on that.

If you'd rather see Android verified on this PR anyway, happy to boot the \`qariah_pixel_7\` AVD and post a follow-up comment — just say the word.

### What was NOT smoke-tested

- Lock-screen metadata sync (RFC-005 PR 5)
- Mushaf ayah advance (RFC-005 PR 4)
- Ambient crossfade (RFC-005 PR 7)
- AppState transitions / interruption recovery

All of these require characterization tests and real-device fidelity per ADR-0001 §How we'll know it worked, and they belong with RFC-005's refactor PRs, not the seam-only RFC.

OKKHALIL3 (approved):

PR #244, opened 2026-05-07
fix: render reciter photos from catalog imageUrl prop

omar-zarka:
## Summary

`ReciterImage` declares an `imageUrl?: string` prop and call sites pass it, but the component's destructure pattern omits it — so the prop is silently dropped and the image is always sourced from the bundled `reciterImages[formattedName]` map (or falls back to `ProfileIcon`).

For reciters whose `formattedName` matches the bundled headshot set, this is masked. For reciters who only have a catalog `image_url` (no bundled asset), the generic `ProfileIcon` shows instead of their photo.

## Changes

- Destructure `imageUrl` in the component signature.
- Compute `imageSource = imageUrl ? {uri: imageUrl} : localImageSource ?? null`.
- Use `imageSource` in the conditional `<Image>` render.
- Pass `recyclingKey={imageUrl || formattedName}` so expo-image's cache invalidates when the URL changes.

## Affected call sites (all already pass `imageUrl`)

- `components/ReciterItem.tsx:51` — `imageUrl={item.image_url || undefined}`
- `components/ReciterCard.tsx:50` — `imageUrl={item.image_url || undefined}`
- `components/ReciterHeader/index.tsx:33,43` — `imageUrl={reciter.image_url || undefined}`

## Behavior

- **Behavior-preserving** for reciters with a bundled headshot whose catalog entry has no `image_url` — local image is still used as the fallback.
- **Behavior-fixing** for reciters whose catalog entry does have an `image_url` — the prop is now actually rendered (previously dropped).

## Test plan

- [ ] Cold-launch app on iOS, scroll the reciters list, verify photos render for reciters whose catalog `image_url` is set but who do not have a bundled local asset.
- [ ] Verify reciters with both a bundled local asset and a catalog `image_url` now show the catalog photo (catalog is the source of truth).
- [ ] Verify `ProfileIcon` fallback still renders for reciters with neither.
- [ ] `npx tsc --noEmit` clean.


osmansaeday:
### Code review

OKKHALIL3 (approved):

osmansaeday (commented):
Follow-up review: this PR updated 10 of 12 `ReciterImage` call sites to pass the new `imageUrl` prop, but two were missed. On iPad these will silently fall back to the bundled local-headshot lookup (or default profile icon) instead of using the catalog `image_url`.

**`components/tablet/TabletPlayerReciterColumn.tsx:228`** — `reciter` is already in scope:

```tsx
<ReciterImage
  imageUrl={reciter.image_url || undefined}
  reciterName={reciter.name}
  style={styles.avatar}
  profileIconSize={moderateScale(18)}
/>
```

**`components/tablet/TabletDockedPlayer.tsx:77` and `:101`** — `currentTrack.artwork` is already populated via `getReciterArtwork(reciter)` in `utils/track.ts`, so it can route through directly:

```tsx
<ReciterImage
  imageUrl={currentTrack.artwork}
  reciterName={currentTrack.reciterName}
  style={styles.artworkCompact}
/>
```

Effect on Bayaan: a reciter added to the catalog with `image_url` populated but no bundled local headshot renders as the default profile icon on the iPad sidebar and docked mini-player. Doesn't repro on phone since FloatingPlayer/MiniPlayer use the patched call sites.

Could you patch in a follow-up?

PR #245, opened 2026-05-07
chore: standardize on tsx for TypeScript script execution

omar-zarka:
## Summary

Replaces `ts-node` with `tsx` for the three `.ts`/`.js` script invocations in `package.json`. `tsx` is an esbuild-based runner that is significantly faster on cold starts, supports both ESM and CJS without configuration, and is the modern default in the Node ecosystem.

### Changes

- `generate-reciter-images`: `ts-node` → `tsx`
- `fetch-reciters`: `ts-node` → `tsx`
- `resize-splash-images`: `ts-node` → `node` (was running a plain `.js` file through `ts-node` unnecessarily)
- devDependency: `ts-node ^10.9.2` → `tsx ^4.19.2`

The two `.ts` scripts use `require()`/CommonJS-style imports today, which `tsx` handles transparently. No script source code changes needed.

## Why upstream

This is a fork-agnostic developer-experience improvement. `tsx` (a) starts faster than `ts-node`, (b) needs no `ts-node` config block in `tsconfig.json`, and (c) handles ESM/CJS/`.mts`/`.cts` interop out of the box. The Qariah fork (qariah-v2) standardized on `tsx` in Sprint 6 (Q2 2026 audit row #92) and confirmed all its existing TypeScript scripts run unmodified.

## Test plan

- [x] `npx tsc --noEmit` clean on this branch
- [x] `npx tsx --version` returns 4.21.0; smoke `console.log` runs
- [ ] Maintainer: confirm `npm run generate-reciter-images` and `npm run fetch-reciters` still work in your environment

## Risk

Low. `tsx` and `ts-node` are API-compatible for the use cases in this repo (CLI invocation only — no `require('ts-node/register')` anywhere). The two `.ts` scripts already use `require()` instead of ESM `import`, so esbuild's transform path is straight-through.

## Cross-ref

This is the upstream side of qariah-v2 divergence-ledger row #92 (Q2 2026 audit candidate E).

PR #247, opened 2026-05-07
chore: add catalog URL verifier with magic-byte sniff

omar-zarka:
## Summary

Adds `scripts/verify-catalog-urls.ts` — a maintainer/CI tool that goes beyond HEAD-200 + Content-Type checks when validating a reciter catalog's audio URLs.

For every `(reciter, rewayat, surah)` entry in the catalog (default: `data/reciters-fallback.json`), the script:

1. Issues a HEAD request and asserts HTTP 200 + `Content-Type: audio/mpeg`.
2. Fetches the first 64 bytes via `Range: bytes=0-63` and asserts MP3 magic bytes (ID3v2 header `ID3` or MPEG audio frame sync `0xFF 0xEx/0xFx`).

The script also recognizes `WAV`, `M4A`, `Ogg`, and `FLAC` magic so failures surface the actual format rather than just "not MP3."

### Flags

- `--catalog <path>` — verify a custom catalog file (default: `data/reciters-fallback.json`)
- `--fast` — skip the default 1 RPS rate-limit (for offline mirrors)
- `--no-sniff` — skip the magic-byte check (HEAD-only diagnostic mode)

A `verify-catalog-urls` script is wired into `package.json` for convenience.

## Why upstream

HEAD-200 + Content-Type checking is necessary but not sufficient. A misconfigured upload pipeline (or a manual S3/R2 bucket put with the wrong source bytes) can place M4A or WAV bytes at a `.mp3` key while the bucket still serves `Content-Type: audio/mpeg`. Every HEAD request returns 200 cleanly, but iOS's audio decoder silently rejects the file at playback time, and the user sees a broken track with no error.

This pattern bit the Qariah fork three times (Sprint 3, Sprint 4, Sprint 5; documented in their `DIVERGENCE_LEDGER.md` rows #51 and #69). The magic-byte sniff is generic infrastructure — useful for any catalog, on any CDN.

## Test plan

- [x] `npm run lint` clean
- [x] Smoke run against `data/reciters-fallback.json`: parses 224 reciters / 268 rewayat / 25208 URLs; first ~20 URLs verified successfully end-to-end with magic-byte sniff (mp3quran.net + R2 sources). Full sweep takes ~7 hr at 1 RPS so abbreviated.

## Risk

Low. New file; zero changes to runtime code or other scripts. The verifier is read-only — never writes to the catalog or any URL.

## Cross-ref

Upstream side of qariah-v2 divergence-ledger row #61 (Q2 2026 audit candidate D).

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "generate-reciter-images": "ts-node scripts/generateReciterImages.ts",
     "generate-icons": "node scripts/generate-icons.js",
     "fetch-reciters": "ts-node scripts/fetchReciterData.ts",
+    "verify-catalog-urls": "ts-node scripts/verify-catalog-urls.ts",
     "resize-splash-images": "ts-node scripts/resize-splash-images.js",
     "prebuild": "LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 expo prebuild --clean",
     "version:patch": "node scripts/version-bump.js patch",
--- a/scripts/verify-catalog-urls.ts
+++ b/scripts/verify-catalog-urls.ts
@@ -106,6 +106,11 @@ function headRequest(
  * Fetches the first 64 bytes of the URL via Range request and returns the
  * detected audio format from magic bytes. Some CDNs ignore Range and return
  * the full body; that's fine — we destroy the response once we have enough.
+ *
+ * Note: `http.request` does not follow 3xx redirects. A catalog URL that
+ * 301/302s to the canonical asset will be reported as a failure here. That
+ * is the intended behavior for a verifier — the catalog should reference
+ * canonical URLs directly.
  */
 function sniffFirstBytes(url: string): Promise<AudioFormat> {
   return new Promise((resolve, reject) => {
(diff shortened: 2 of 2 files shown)


osmansaeday:
Code review is clean — magic-byte detection covers all 6 formats correctly (ID3v2, MPEG sync 0xFF/0xEx-0xFx, RIFF/WAVE, ftyp, OggS, fLaC), no new deps, rate-limited by default, read-only. One coordination call before merging:

The new `package.json` entry uses `ts-node`:
```json
"verify-catalog-urls": "ts-node scripts/verify-catalog-urls.ts",
```
…but #245 (still open) removes `ts-node` from devDeps entirely. Whichever lands second will need the other's update. Suggest landing this PR first, then folding the `verify-catalog-urls` line into #245's tsx migration.

Two tiny non-blocking notes for a follow-up if you ever revisit:
1. `sniffFirstBytes` registers both `'end'` and `'close'` resolvers — after `res.destroy()` both fire. Promise eats the second resolve so it's safe, just slightly unclean. Either guard with a `done` flag or rely only on `'close'`.
2. `http.request` doesn't follow redirects by default, so any catalog URL that 301/302s will be reported as a failure. Probably the *correct* behavior for a verifier (you want direct URLs) but worth knowing.

Otherwise looks good — happy to merge once the #245 coordination is decided. Branch will need `gh pr update-branch --rebase` first since `develop` has moved on.

omar-zarka:
Thanks for the review! Pushed `04efa9d` addressing both notes:

1. **`sniffFirstBytes` resolver** — dropped the `'end'` handler; `'close'` fires on both natural end-of-stream and after `res.destroy()`, so it's a strict superset. Single resolver path now.
2. **Redirect note** — added a JSDoc paragraph above `sniffFirstBytes` explaining `http.request` doesn't follow 3xx and that this is intentional for a verifier.

Branch was already in sync with `origin/develop` — no rebase needed. CI should fire on this push.

On the #245 coordination: agreed, let's land #247 first. Once this merges I'll rebase #245 against the new `develop` and fold the `verify-catalog-urls` line into the tsx migration there.

PR #250, opened 2026-05-08
chore(ios): allow Sentry source-map upload to fail soft during archive

osmansaeday:
## Summary

Set `SENTRY_ALLOW_FAILURE=true` at the top of `scripts/ios-archive.sh` so transient network errors during the Sentry CLI source-map upload (DNS, SSL_read, etc) no longer kill a release archive.

## Context

Hit this twice today during a TestFlight build for `feature/qcf-v2-mushaf`:

1. ``error: API request failed → Could not resolve host: sentry.io``
2. ``error: API request failed → OpenSSL SSL_read: unexpected eof while reading``

In both cases the JS bundle was fully built; only the Sentry source-map upload step failed, and it took the whole archive down (`PhaseScriptExecution Bundle\ React\ Native\ code\ and\ images` exited 1 → `** ARCHIVE FAILED **`).

`SENTRY_ALLOW_FAILURE=true` is the documented escape hatch (the `sentry-cli` error message itself recommends it). Symbolication is best-effort; missed source maps can be uploaded out-of-band via `sentry-cli sourcemaps upload` later.

## Test plan

- [ ] Run `npm run ios:archive` while online: archive succeeds, source maps still upload
- [ ] Run `npm run ios:archive` with sentry.io blocked (e.g. `127.0.0.1 sentry.io` in `/etc/hosts`): archive still succeeds; source-map upload prints an error but does not abort the build
- [ ] Pull the build into TestFlight and confirm the IPA installs and runs as expected


OKKHALIL3:
Makes sense as a pragmatic fix — blocking a release archive on a transient DNS/SSL error is the wrong trade-off, and `SENTRY_ALLOW_FAILURE` is the right lever for this.

One thing worth flagging: with this in place, source-map upload failures become silent. If a non-transient issue starts causing uploads to fail consistently (wrong token, changed endpoint, etc.), we won't know — crash reports on Sentry will quietly lose symbolication and we'd only notice when trying to debug a production incident.

Would it be worth adding a post-upload check or at minimum logging the sentry-cli exit code so we at least get a warning in the build output, even if we don't fail? That way the archive still succeeds but someone skimming the build log would catch a pattern of failures before it becomes a problem.

PR #251, opened 2026-05-09
chore: add verify-version-sync.js pre-archive guard (with --fix flag)

omar-zarka:
## Summary

Adds `scripts/verify-version-sync.js` — a pre-archive sanity check that the version + build values in `app.config.ts` (via `generate-version.js`), the iOS `Info.plist`, and `android/app/build.gradle` all agree. Wired into `scripts/ios-archive.sh` (shipped via #241) so the archive helper auto-patches the working tree on drift instead of failing fast.

## Why

`expo prebuild` regenerates the native dirs from `app.config.ts`, but if you bump the config and forget to re-prebuild (or commit the prebuild output to a separate branch), the archive ships with stale version metadata:

- Apple rejects uploads whose `CFBundleShortVersionString` is below a previously-shipped value.
- Play Store rejects builds with a stale `versionCode`.

Catching this at archive time (after a 5–10 minute build) costs a full extra build cycle. The pre-flight check costs ~50 ms.

I hit this twice during my Qariah fork's launch push: once when the v1 Flutter app's `3.0.1` history forced my v2 fork from `2.2.0` → `3.1.0`, and again on the next archive when committing the version-sync bumped the git-rev-list count by 1 and left the native files 1 behind the new source-of-truth — a moving-target failure that fails the very check I'd just done.

`--fix` solves the second pattern: auto-patches the working tree (`ios/<App>/Info.plist` + `android/app/build.gradle`) to match `generate-version.js`, re-verifies post-patch, exits 0. Idempotent. Standing rule: archive BEFORE committing the resulting sync.

## Behavior

```
node scripts/verify-version-sync.js          # check-only; exit 1 on drift
node scripts/verify-version-sync.js --fix    # auto-patch native files, re-verify

# Exit codes:
#   0 — all 3 in sync (after --fix, means the patch succeeded)
#   1 — mismatch (without --fix; with --fix, only if patch failed)
#   2 — parse error in one of the inputs
```

Fork-agnostic: scans `ios/<App>/Info.plist` rather than hardcoding any particular app name (mirrors the workspace auto-detect already shipped in `scripts/ios-archive.sh` via #241). Drop-in for any RN-Expo project that uses `generate-version.js`.

## Test plan

- [x] No-drift baseline: exits 0, prints "All three sources agree."
- [x] Drift detected without `--fix`: exits 1, prints diagnostic + suggested `sed` commands + hint to use `--fix`.
- [x] Drift detected with `--fix`: patches both files, re-verifies, exits 0.
- [x] Restoration: reverting backups via `cp` and re-running shows clean state again.
- [x] Wired into `scripts/ios-archive.sh` as the first step after the `ExportOptions.plist` pre-flight; archive run that previously failed at version-mismatch now self-heals and proceeds.

## Companion to

- #241 (fork-agnostic `ios-archive.sh`) — this PR adds a pre-flight to that script.
- Mirrors the spirit of #239's bootstrap-android-local-properties.js postinstall guard: catch a class of "I forgot to run prebuild" failures earlier and cheaper.


OKKHALIL3 (changes requested):
The intent here is solid — version drift causing App Store rejections is a real problem worth guarding against. A few issues before I can approve though.

**Bug: `findInfoPlistPath()` can patch the wrong file**

`fs.readdirSync` returns entries in filesystem order, and `ios/` typically contains test target directories (`BayaanTests`, `BayaanUITests`) that also have `Info.plist` files. If one of those sorts before the app directory, `--fix` patches a test target's plist instead of the app's — the exact problem this script was meant to prevent. The discovery logic needs to either filter out known non-app directories or key off the app name directly.

**Bug: `generate-version.js` stdout parsing is fragile**

```js
const m = out.match(/\{[\s\S]*?\}/);
const json = JSON.parse(m[0]);
```

Non-greedy `*?` grabs the first `{...}` in stdout. If `generate-version.js` ever emits any log line before the JSON, this silently parses the wrong object — exit code stays 0 and the script proceeds with garbage version values. Better to write the version to a temp JSON file and read that, rather than parsing stdout.

**Bug: the suggested `sed` commands in error mode can corrupt the plist**

The script suggests `sed -i '' 's|<string>1.0.0</string>|...'` but plists often contain the version string in multiple keys. That sed replaces all occurrences, not just `CFBundleShortVersionString`. A developer following the script's own error output could silently corrupt the plist.

**Design: `--fix` always mutates the working tree**

The archive script calls `--fix` unconditionally, so every archive run potentially writes to native files regardless of whether there's drift. The "archive before committing" rule is buried in a comment with no enforcement — if someone commits first the guard is defeated. Worth making `--fix` opt-in and having the default mode exit non-zero on drift so the developer consciously decides to apply it.

**Design: Android check is embedded in the iOS archive with no skip**

If Android and iOS are on different release schedules, a legitimate Android build number difference will block an iOS archive with no escape hatch. A `--platform ios` flag or similar would help.

**No tests for the patching logic**

`patchInfoPlist` and `patchAndroidGradle` write to native files that directly affect App Store submissions. Given the bugs already present in the discovery and parsing code, I'd want at least basic unit tests for the patch functions before this touches a release build.

osmansaeday:
### Code review

Found 1 issue:

1. `patchInfoPlist` writes a literal tab (`\n\t<string>`) for the indent before `<string>`, but the existing `ios/Bayaan/Info.plist` is 4-space-indented. Every `--fix` run will swap 4 spaces for a tab on the two version lines, polluting the diff with whitespace churn so the post-fix file never matches the rest of the plist's indentation. Confirmed with `od -c` on `ios/Bayaan/Info.plist`: the lines preceding `CFBundleShortVersionString` and `CFBundleVersion` start with 4 space chars, not a tab. Suggest reading the existing line's leading whitespace and reusing it, or hardcoding `'    '` (4 spaces) to match expo-prebuild's output.

https://github.com/thebayaan/Bayaan/blob/9b74915b3d6612b0bfdf3d3c54342df4309c4534/scripts/verify-version-sync.js#L114-L129

Compare to the indentation in the file being patched:

https://github.com/thebayaan/Bayaan/blob/28259518dd09fb1c14e85f303f81f069d3f60f24/ios/Bayaan/Info.plist#L25-L48

omar-zarka:
Thanks for the careful review @OKKHALIL3. Each issue addressed in its own commit so the mapping is easy to follow:

| # | Concern | Commit |
|---|---|---|
| 1 | `findInfoPlistPath()` could patch wrong target | `aa4324c` — filter `*Tests`/`*UITests`/`*Extension`/`*Widget` suffixes + `Pods`/`build`/`DerivedData` + `.xcodeproj`/`.xcworkspace`. If multiple candidates remain after filtering, throw with the list rather than picking arbitrarily. |
| 2 | Fragile stdout regex parsing | `7613bf0` — added `--json-only` flag to `generate-version.js`; verify-version-sync.js calls with that flag and `JSON.parse(stdout)` directly. |
| 3 | `sed` suggestion can corrupt | `63bdade` — dropped the sed block. Mismatch output now just shows "has X, needs Y" + points to `--fix`. Comment in the script explains why no manual command is offered. |
| 4 | `--fix` always mutates | `939f063` — `ios-archive.sh` runs check-only by default. On drift, prints a clear "re-run with `VERIFY_FIX=1 ${0}`" message. Auto-fix is opt-in per archive run. |
| 5 | No `--platform` skip | `17e5211` — added `--platform=ios\|android\|both` (default both for direct invocation). `ios-archive.sh` passes `--platform=ios`. Skipped platform's check is bypassed entirely (no parser run, no compare). |
| 6 | No tests for patch logic | `bcad9e4` — added `scripts/__tests__/verify-version-sync.test.js` with 21 tests using `os.tmpdir()` fixtures. Covers discovery (incl. the bug-1 regression), parsers, patchers (incl. the bug-3 same-value collision regression), and round-trip. Required exporting the pure functions from `verify-version-sync.js`. **Suite: 90 → 111 green.**

Re-review when you have time.

omar-zarka:
Thanks for the catch — but `od -c` on the actual file shows tab-indented, not 4-space:

```
$ sed -n '25,30p' ios/Bayaan/Info.plist | od -c | head -4
0000000   \t   <   k   e   y   >   C   F   B   u   n   d   l   e   P   a
0000020    c   k   a   g   e   T   y   p   e   <   /   k   e   y   >  \n
0000040   \t   <   s   t   r   i   n   g   >   $   (   P   R   O   D   U
```

Each line preceding `<key>CFBundleShortVersionString</key>` and `<key>CFBundleVersion</key>` starts with `\t`, matching what `expo prebuild` emits. So `\n\t<string>${target.semanticVersion}</string>` produces output identical to the surrounding indentation — `--fix` runs are clean (no whitespace churn in the diff). I just verified end-to-end by running `--fix` against a drift state and the only changed bytes were the version values themselves.

Worth noting if a future fork's `expo prebuild` ever emits 4-space-indented Info.plist (some plugin chains do), reading the existing line's leading whitespace would be the more robust play. Tracked but not changing for this PR since current behavior is correct against the actual file. Happy to revisit if you've seen 4-space indent elsewhere.

omar-zarka:
Thanks for the careful review — all six points addressed in 7 follow-up commits. Reverse map:

| Your concern | Fix | Commit |
|---|---|---|
| Bug 1: `findInfoPlistPath()` patches test-target Info.plist when `BayaanTests/`/`BayaanUITests/` sort first | Added an `IOS_NON_APP_SUFFIXES` denylist (`Tests`, `UITests`, `Snapshots`, etc.) + skip `Pods/`, `build/`, `DerivedData/`. Picks the first remaining directory containing an `Info.plist`. | `aa4324c` |
| Bug 2: `generate-version.js` stdout regex `match(/\{[\s\S]*?\}/)` grabs garbage if a log line precedes | Gave `generate-version.js` a `--json-only` flag that writes only the JSON document. Verify script invokes with that flag and `JSON.parse(out)` directly — no regex. | `7613bf0` |
| Bug 3: suggested `sed -i '' 's\|<string>1.0.0</string>\|...'` corrupts plists (multiple keys match) | Dropped the sed suggestions entirely. Mismatch mode now prints the structured diff and exits non-zero — developer chooses how to remediate. | `63bdade` |
| Design 4: `--fix` always mutates working tree from `ios-archive.sh` | `--fix` is now opt-in via `VERIFY_FIX=1` env var. Default mode exits non-zero on drift; the archive script's verify step also defaults to verify-only. | `939f063` |
| Design 5: Android check embedded in iOS archive with no escape hatch | Added `--platform=ios\|android\|both` flag (default `both`). iOS archive script passes `--platform=ios`. | `17e5211` |
| Design 6: no tests for `patchInfoPlist` / `patchAndroidGradle` | Added `scripts/__tests__/verify-version-sync.test.js` — 21 tests covering parsers (Info.plist + gradle), `findInfoPlistPath` filtering, both patchers (no-op when equal, surgical when one differs, round-trip), and version comparator. | `bcad9e4` |
| (housekeeping) prettier | Auto-fix applied. | `93dbd5f` |

Validation locally:
- `npx jest scripts/__tests__/verify-version-sync.test.js` — **21 passed**
- `npx eslint scripts/verify-version-sync.js scripts/generate-version.js` — clean

Re-requesting review when you have a moment. Happy to iterate further.

omar-zarka:
Hi @OKKHALIL3 — friendly nudge on this one: all six review points addressed across the follow-up commits (filter non-app dirs, `--json-only` mode for `generate-version.js`, dropped the unsafe sed suggestions, opt-in `--fix` via `VERIFY_FIX=1`, `--platform=ios|android|both`, 21 unit tests). Ready for re-review whenever you have a moment.

While I'm here, in case it helps your queue planning, the Qariah-side queue has built up a bit — happy to land these in any order or sequencing that works best for you:

| PR | Status | Notes |
|---|---|---|
| [#244](https://github.com/thebayaan/Bayaan/pull/244) | ReciterImage prop-drop fix | Pure bug fix, single-file, no API change. Same drop pattern as #252 (already merged). |
| [#245](https://github.com/thebayaan/Bayaan/pull/245) | `tsx` over `ts-node` | Chore. You commented; if there's a remaining concern happy to dig in. |
| **#251** | This PR | Re-review pending after CHANGES_REQUESTED resolution. |
| [#255](https://github.com/thebayaan/Bayaan/pull/255) | analytics `reciter_name` | Schema-only addition; 31 tests pass. |
| [#260](https://github.com/thebayaan/Bayaan/pull/260) | `branding.homeRowConfig` (RFC-008/XF-NNN) | Code PR. Default preserves existing Bayaan row order. |
| [#261](https://github.com/thebayaan/Bayaan/pull/261) | RFC-008 doc — `branding.listenTabTopComponent` | Doc-only; code-PR pre-staged for once the shape clears. |
| [#262](https://github.com/thebayaan/Bayaan/pull/262) | RFC-009 doc — translation/tafsir provider seams | Doc-only. |
| [#263](https://github.com/thebayaan/Bayaan/pull/263) | `branding.catalogVersionEndpoint` (RFC-010) | Code + 5 unit tests. |
| [#264](https://github.com/thebayaan/Bayaan/pull/264) | RFC-011 doc — `branding.continueReadingHistorySize` | Doc-only. |

No urgency — just a single round-up rather than nine separate pings. Thanks as always for the bandwidth you give these.

OKKHALIL3 (approved):
Jazak Allahu Khairun

PR #252, opened 2026-05-09
fix: render FollowAlongBadge on Browse + Circular reciter cards

omar-zarka:
## Summary

Both `components/browse/BrowseReciterCard.tsx` and `components/cards/CircularReciterCard.tsx` declare `showFollowAlong?: boolean` in their props interface but never destructure or render it. When a caller passes `showFollowAlong={true}` (e.g. via the `useReciterFollowAlong` hook surfaced through `RecitersView.tsx`), the badge silently disappears on both card variants — only the list-row variant (`components/ReciterItem.tsx`) renders it.

Same shape as the `imageUrl` prop drop fixed in #244.

## Changes

- `components/browse/BrowseReciterCard.tsx`: destructure `showFollowAlong`; render `<FollowAlongBadge />` in a new flex row alongside the rewayat label. Added `showFollowAlong` to the `React.memo` comparison so toggles re-render.
- `components/cards/CircularReciterCard.tsx`: destructure `showFollowAlong`; render `<FollowAlongBadge />` in the subtitle slot when true, replacing the static "Reciter" label.

## How I found it

`react/no-unused-prop-types` ESLint rule (landed via #248) — both cards showed up as findings. 2 real bugs in the inherited 236-finding count. The remaining ones are mostly RN `renderItem` `ListRenderItemInfo<T>` false positives and some IconProps shared-interface stylistic choices.

## Test plan

- [x] `tsc --noEmit` clean
- [x] Tested on Qariah fork with `useReciterFollowAlong(reciterId)` returning true; badge now renders on both card variants in the home grid + reciter carousel.
- [ ] Visual regression: when `showFollowAlong={false}` (default), card layout is unchanged.

### DIFF
--- a/components/browse/BrowseReciterCard.tsx
+++ b/components/browse/BrowseReciterCard.tsx
@@ -4,6 +4,7 @@ import {moderateScale} from 'react-native-size-matters';
 import {Reciter} from '@/data/reciterData';
 import {Theme} from '@/utils/themeUtils';
 import {ReciterImage} from '@/components/ReciterImage';
+import {FollowAlongBadge} from '@/components/badges/FollowAlongBadge';
 import Color from 'color';
 import {GlassView} from 'expo-glass-effect';
 import {Link} from 'expo-router';
@@ -60,6 +61,12 @@ function createStyles(theme: Theme, width: number, height: number) {
       color: Color(theme.colors.textSecondary).alpha(0.5).toString(),
       lineHeight: moderateScale(12),
       includeFontPadding: false,
+      flexShrink: 1,
+    },
+    secondaryRow: {
+      flexDirection: 'row',
+      alignItems: 'center',
+      gap: moderateScale(4),
     },
     glassContainer: {
       borderWidth: 0,
@@ -80,6 +87,7 @@ const BrowseReciterCard = React.memo(
     height,
     theme,
     rewayatId,
+    showFollowAlong,
   }: BrowseReciterCardProps) => {
     const glassColorScheme = useGlassColorScheme();
     const styles = useMemo(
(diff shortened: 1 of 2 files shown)


OKKHALIL3 (approved):

osmansaeday (commented):
Follow-up review: the PR title says "render FollowAlongBadge on Browse + Circular reciter cards", and the patch correctly threads `showFollowAlong` through the new prop on `BrowseReciterCard`. But the Browse-tab render path goes through `BrowseGrid`, which neither accepts nor forwards the prop, so on Browse the badge never appears today.

Wiring on `develop`:

```
BrowseReciters → BrowseGrid → BrowseReciterCard
                  (no showFollowAlong) (showFollowAlong is always undefined)
```

`components/browse/BrowseGrid.tsx` (the `<BrowseReciterCard>` call site around L160) passes `key`, `reciter`, `onPress`, `width`, `height`, `theme`, `rewayatId` but not `showFollowAlong`. The `BrowseGridProps` interface (L13) also omits it.

`RecitersView` works (it calls `useReciterFollowAlong(reciterId)` once per row and threads it down to `ReciterItem` / `CircularReciterCard`), so the badge does render on the Listen tab. Only the Browse-tab path is silent.

Two clean fixes:

**Option A** (smallest patch): have `BrowseReciterCard` call the hook internally and drop the prop entirely.

```tsx
// inside BrowseReciterCard
const showFollowAlong = useReciterFollowAlong(reciter.id);
```

Pro: zero caller changes. Con: hook fires once per card, but the selector is cheap (a `Set.has` against `useTimestampStore`) and React-deduped, so it's fine.

**Option B**: pass a precomputed `Set<string>` of follow-along reciter ids through `BrowseGrid`:

```tsx
// BrowseReciters
const followAlongIds = useTimestampStore(s => s.supportedReciterIds);

<BrowseGrid ... followAlongIds={followAlongIds} />

// BrowseGrid
<BrowseReciterCard ... showFollowAlong={followAlongIds.has(reciter.id)} />
```

Pro: one store subscription for the whole grid. Con: needs a new prop on `BrowseGridProps`.

A is probably the right call given the call-once-per-row cost is negligible. Could you patch in a follow-up?

PR #254, opened 2026-05-09
feat(branding): tab label overrides for fork customization

omar-zarka:
## Summary

Adds an optional `tabs` namespace to `config/branding` so a fork can rename the bottom tabs (Home / Surahs / Search / Collection / Settings) without patching `app/(tabs)/_layout.tsx`. Today these labels are hardcoded in three places (iOS NativeTabs, Android `Tabs.Screen`, tablet sidebar `Tabs.Screen`) and any rename produces an 8-line diff that conflicts on every upstream merge.

Extends RFC-007 (App-layer multi-tenancy seams). Zero behavior change for Bayaan.

## Motivation

RFC-007 covered identity strings (`appName`, support URLs, etc.) but didn't extend to in-app navigation copy. Forks regularly want different framing — Qariah is renaming "Home" to "Listen" (audio-focused product) and "Surahs" to "Mushaf" (Arabic-first audience), and other forks have asked about similar tweaks.

The current alternative is a local patch to 15 string literals across `_layout.tsx` (5 in `IOSTabs`, 5 in `AndroidTabs`, 5 in `TabletTabs`). And `_layout.tsx` is exactly the file that gets touched whenever upstream adds a tab feature (mini-player, tablet sidebar, etc.), maximizing conflict risk.

The override is purely additive — every label falls back to Bayaan's current string when the config key is absent.

## Scope: labels only, not icons

This PR intentionally **does not** add icon overrides. Bayaan's tab icons are wired through three different surfaces:

- iOS NativeTabs: PNG `require()` calls bundled at compile time (Metro static-analysis requirement).
- Android `BottomTabBar`: switch on `routeName` returning React component (`HomeIcon`, `QuranIcon`, …).
- Tablet sidebar: same switch as Android.

Refactoring to a fork-overridable icon registry while preserving NativeTabs' template rendering and Metro's static `require()` discovery is a substantially larger change that would benefit from its own RFC. Forks needing a different icon (Qariah wants headphones for the renamed "Listen" tab) can do a small local edit to the three switch cases — much narrower than the 15-string label changes this PR addresses.

## Design

`config/branding.d.ts`:

```ts
export interface BrandingTab {
  label?: string;
}

export interface BrandingTabs {
  home?: BrandingTab;
  surahs?: BrandingTab;
  search?: BrandingTab;
  collection?: BrandingTab;
  settings?: BrandingTab;
}

export interface Branding {
  // …existing fields…
  tabs?: BrandingTabs;  // optional; absent on Bayaan
}
```

`app/(tabs)/_layout.tsx`:

```ts
import branding from '@/config/branding';

const tabLabels = {
  home: branding.tabs?.home?.label ?? 'Home',
  surahs: branding.tabs?.surahs?.label ?? 'Surahs',
  search: branding.tabs?.search?.label ?? 'Search',
  collection: branding.tabs?.collection?.label ?? 'Collection',
  settings: branding.tabs?.settings?.label ?? 'Settings',
};

// IOSTabs:   <NativeTabs.Trigger.Label>{tabLabels.home}</NativeTabs.Trigger.Label>
// AndroidTabs / TabletTabs:   <Tabs.Screen … options={{title: tabLabels.home}} />
```

That's the entire diff: one import, one constant table, and search-and-replace of 15 string literals across three layout components. `BottomTabBar` and `TabletSidebar` both already read `options.title` from React Navigation, so they pick up the override automatically without further changes.

## Why labels-only and not a full tab registry

A fully generic tab system (`branding.tabs = [...]` with arbitrary route names, ordering, hide/show flags) sounds tempting but blows up scope:

- Route names are filesystem-bound (`app/(tabs)/(a.home)` etc.); reordering or hiding requires Expo Router config + screen reachability handling.
- `BottomTabBar`'s icon switch + navigation hooks would need to become dynamic.
- Tablet sidebar has additional layout logic (rail vs. labeled, mini-player dock).

**Labels-only is the 80/20** — covers every fork-renaming use case I've heard, with a ~60-line diff and zero structural change. If/when there's clear demand for tab reordering or hide flags, that's a separate RFC.

## Backward compatibility

- `branding.tabs` is optional; absent on Bayaan today.
- Each `tabs.<name>.label` is optional; nullish-coalesces to the current hardcoded string.
- No type breaks, no runtime breaks, no asset changes.
- Tablet sidebar inherits via React Navigation `options.title` — no separate override path needed.

## Test plan

- [x] `tsc --noEmit` clean locally.
- [ ] CI lint + tests pass.
- [ ] Bayaan boot-gate (iOS + Android + iPad): all five tab labels render as "Home / Surahs / Search / Collection / Settings" — visual regression check.
- [ ] Manual smoke: temporarily add `tabs: {home: {label: 'TEST'}}` to `config/branding.js`, confirm label updates on iOS NativeTabs, Android BottomTabBar, and tablet sidebar. Verify removal restores normal appearance.

## Companion fork PR

Qariah will land a no-op shim of this change immediately after merge: removes its local `QARIAH-PENDING-UPSTREAM` patch and starts using `tabs: {home: {label: 'Listen'}, surahs: {label: 'Mushaf'}}` directly in its `config/branding.js`.

### DIFF
--- a/app/(tabs)/_layout.tsx
+++ b/app/(tabs)/_layout.tsx
@@ -14,6 +14,17 @@ import {FloatingPlayer} from '@/components/player/v2/FloatingPlayer';
 import {TabletSidebar} from '@/components/tablet/TabletSidebar';
 import {useResponsive} from '@/hooks/useResponsive';
 import {USE_GLASS} from '@/hooks/useGlassProps';
+import branding from '@/config/branding';
+
+// Resolved once at module load. Bayaan's defaults are used when a fork
+// hasn't supplied a `branding.tabs.<name>.label` override.
+const tabLabels = {
+  home: branding.tabs?.home?.label ?? 'Home',
+  surahs: branding.tabs?.surahs?.label ?? 'Surahs',
+  search: branding.tabs?.search?.label ?? 'Search',
+  collection: branding.tabs?.collection?.label ?? 'Collection',
+  settings: branding.tabs?.settings?.label ?? 'Settings',
+};
 
 // PNG tab icons for NativeTabs (iOS) — template-rendered by the native tab bar
 const tabIcons = {
@@ -66,7 +77,7 @@ function IOSTabs() {
         name="(a.home)"
         contentStyle={{backgroundColor: theme.colors.background}}>
         <NativeTabs.Trigger.Icon src={tabIcons.home} renderingMode="template" />
-        <NativeTabs.Trigger.Label>Home</NativeTabs.Trigger.Label>
+        <NativeTabs.Trigger.Label>{tabLabels.home}</NativeTabs.Trigger.Label>
       </NativeTabs.Trigger>
 
       <NativeTabs.Trigger
@@ -76,7 +87,7 @@ function IOSTabs() {
           src={tabIcons.surahs}
           renderingMode="template"
         />
-        <NativeTabs.Trigger.Label>Surahs</NativeTabs.Trigger.Label>
+        <NativeTabs.Trigger.Label>{tabLabels.surahs}</NativeTabs.Trigger.Label>
       </NativeTabs.Trigger>
 
       <NativeTabs.Trigger
@@ -87,7 +98,7 @@ function IOSTabs() {
           src={tabIcons.search}
           renderingMode="template"
         />
-        <NativeTabs.Trigger.Label>Search</NativeTabs.Trigger.Label>
+        <NativeTabs.Trigger.Label>{tabLabels.search}</NativeTabs.Trigger.Label>
       </NativeTabs.Trigger>
 
       <NativeTabs.Trigger
@@ -97,7 +108,7 @@ function IOSTabs() {
           src={tabIcons.collection}
           renderingMode="template"
         />
-        <NativeTabs.Trigger.Label>Collection</NativeTabs.Trigger.Label>
+        <NativeTabs.Trigger.Label>{tabLabels.collection}</NativeTabs.Trigger.Label>
       </NativeTabs.Trigger>
 
       <NativeTabs.Trigger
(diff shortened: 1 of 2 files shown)


OKKHALIL3 (changes requested):
Holding on this one for a few reasons.

**1. Half-built abstraction — labels without icons**

The code comment in `branding.d.ts` explicitly says icon overrides are deferred to a future RFC. So right now a fork can rename "Surahs" to "Chapters" but can't change the icon sitting next to it. That's an inconsistent seam — a fork using this system ships a mismatched label/icon pair until icons land. I'd rather these ship together in one RFC than have a partial system that forks depend on before it's coherent.

**2. Tab labels resolved at module load**

```ts
const tabLabels = {
  home: branding.tabs?.home?.label ?? 'Home',
  ...
};
```

This bakes labels at module import time. If we ever add runtime language switching or dynamic branding this pattern blocks it. It's a subtle constraint we're committing to now that forecloses options later.

**3. No RFC doc**

Same concern as #253 — this is architectural work that establishes an API surface forks will depend on. It needs a spec. What does the full `tabs` branding system look like? What's in scope and explicitly out of scope? Without that I can't tell if this is the right slice or a premature one.

omar-zarka:
Thanks @OKKHALIL3 — withdrawing this alongside #253 for the same reason and I think your critique here is sharper than mine was when I scoped this PR.

You're right that "labels only, icons deferred to a future RFC" is exactly the kind of half-built abstraction that becomes load-bearing before the full system is coherent. A fork that adopts this would ship mismatched label/icon pairs until the icon RFC landed — which is the worst of both worlds for users of the abstraction.

The module-load resolution critique is also fair. I picked it for simplicity but it forecloses runtime language switching and dynamic branding for everyone, not just us, and that's not a constraint Bayaan should commit to lightly.

Same plan as #253: this collapses into **RFC-008: Branding theme + tabs as a coherent multi-tenant surface**. The icon-overrides design is the open question — needs a registry pattern that survives Metro static `require()` analysis. We'll work that out in the RFC before any code.

Full internal retro at `planning/sprint-10-11-learnings.md` on the Qariah side. Thanks for holding the line on this — the platform extraction is too important to compromise with half-built seams.

PR #260, opened 2026-05-15
feat: branding.homeRowConfig — multi-tenant Listen-tab row order

omar-zarka:
## Summary

Adds an optional `homeRowConfig: HomeRow[]` field to the `Branding` interface, letting forks reorder, hide, or omit Listen-tab home rows without forking `RecitersView.tsx`. **Bayaan behavior is unchanged** — the field has a built-in default that mirrors today's hardcoded order, and the new code path is opt-in.

Continues the multi-tenancy work that started with [RFC-007](https://github.com/thebayaan/Bayaan/pull/235) (which shipped `Branding.appName`, `featureFlags`, `CatalogProvider`).

## What this changes

**`config/branding.d.ts`**
- New `HomeRowId` union (12 ids — one per existing section in `RecitersView`).
- New `HomeRow` interface (`{id: HomeRowId, enabled: boolean}`).
- New optional `Branding.homeRowConfig?: HomeRow[]` field.

**`config/branding.js`**
- Explicit `homeRowConfig` array with all 12 rows enabled in today's render order. Acts as both runtime default for Bayaan and a documentation example for forks.

**`components/RecitersView.tsx`**
- Replaces the inline ``{cond && <Section …/>}`` block with a `rowNodes: Record<HomeRowId, React.ReactNode>` registry + a config-driven `.filter(enabled).map(render)` block.
- Each registry entry preserves the existing conditional gating verbatim (e.g. ``validRecentTracks.length > 0 ? <Section …/> : null``); runtime semantics are identical.
- Falls back to a built-in `DEFAULT_HOME_ROW_CONFIG` when `branding.homeRowConfig` is undefined (defensive — Bayaan's `branding.js` always sets it, but forks may omit).
- Hoisted Adhkar's inline `renderItem` + `keyExtractor` to module-level helpers to avoid `react/no-unstable-nested-components`.

## Why pre-built `ReactNode` (vs factory functions)

I tried `Record<HomeRowId, () => React.ReactNode>` first — cleaner-looking but trips `react/no-unstable-nested-components` because the factories close over hooks-derived state. Pre-built JSX has identical runtime cost (each branch's `null` is what the current `&&` chain already does) and keeps the lint clean.

## What about empty-data rows?

Two-layer hiding: a row hides if `enabled: false` in the config OR if its underlying data is empty (existing behavior preserved per-entry). Forks can leave all rows `enabled: true` and rely on data-empty self-hide for "show when relevant" semantics, or set `enabled: false` to hard-hide a row even when data exists.

## What about unknown ids?

Silently skipped. Forward-compat for forks that haven't been updated when a new upstream row is added — they continue to work, just without the new row.

## Verification

- `npx tsc --noEmit` — clean
- `npx eslint components/RecitersView.tsx config/branding.{js,d.ts}` — no new findings (2 pre-existing warnings unchanged: ``verticalScale`` import, ``theme`` prop on MemoizedFlatList)
- Manual: same 12 rows render in same order when `homeRowConfig` is at its default

## Follow-on RFC (for context, not in this PR)

A companion XF-NNN RFC for the Listen-tab top hero region (`branding.listenTabTopComponent`) will open separately as a doc-only PR first — same fork-need, different abstraction (top tile shape vs row order).

## Test plan

- [ ] Pull this branch, `npm install`, run the app on iOS or Android.
- [ ] Confirm the Listen tab renders identically to `develop` (same 12 rows, same order, same conditional hiding when data is empty).
- [ ] Optional: temporarily edit `config/branding.js` `homeRowConfig` to reorder/disable rows and confirm the UI follows.


osmansaeday:
### Code review

osmansaeday (approved):
LGTM

PR #261, opened 2026-05-15
docs: RFC-008 — listen-tab top-region component slot (doc-only)

omar-zarka:
## Summary

Proposes an opt-in `Branding.listenTabTopComponent?: ComponentType` slot for the Listen tab's top region (currently the hardcoded `<RecitersHero />` in `components/RecitersView.tsx`). Bayaan's behavior is unchanged when the slot is empty.

**Doc-only by design.** The component-as-config pattern is unusual for a `Branding` interface — opening this as a design discussion first lets us align on the shape before any code lands. The RFC explicitly considers three alternatives (declarative tile schema, string enum, status quo) and explains why a plain component slot won.

If the maintainer prefers an alternative shape, this RFC is withdrawn and a new one opens with the preferred shape.

## Why this RFC

This is the second in the XF-NNN multi-tenancy series:
- **RFC-007** (merged): the `Branding` interface itself
- **PR #260** (open as of this writing): `branding.homeRowConfig` for the rows beneath the hero
- **RFC-008** (this PR, doc-only): the hero region

After this RFC + a follow-on code PR land, Qariah's diff for `components/RecitersView.tsx` drops from ~620 lines (its current shared-file divergence) to ~10 lines (the title-string override that PR #260 introduces).

## What's in the RFC

The full doc lives at `docs/rfcs/008-listen-tab-top-slot.md`:
- **Motivation:** real cite of where `RecitersHero` is hardcoded today + the Qariah-side workaround.
- **Decision:** the proposed `ComponentType` slot + the ~10-line Bayaan-side change.
- **Alternatives considered:** declarative tile schema, string enum, status quo — each rejected with reasoning.
- **Consequences:** positive / neutral / negative with the unusual-pattern caveat called out explicitly.
- **Open questions:** three deliberate non-decisions (slot props shape, error boundary, footer slot).

## What this PR is NOT

- **Not a code PR.** No `RecitersView.tsx` changes, no `branding.{d.ts,js}` changes. Just the RFC doc.
- **Not a commitment to land the code as-described.** If the design discussion produces a better shape, the code PR follows the agreed shape.

## Test plan

- [ ] Read `docs/rfcs/008-listen-tab-top-slot.md`.
- [ ] Comment with maintainer preference: ship as proposed / pick an alternative / reject.


osmansaeday (approved):
LGTM

PR #262, opened 2026-05-17
docs(rfc): RFC-009 — external content provider seams (translations + tafsir) (doc-only)

omar-zarka:
**Updated 2026-05-16:** scope broadened to include tafsir after maintainer-side discussion — the two services are structurally identical and benefit from being introduced together.

## What

Doc-only RFC. Proposes two optional `Branding` slots — `translationProvider?: TranslationProvider` and `tafsirProvider?: TafsirProvider` — so forks can supply their own translation and tafsir sources for the Settings → Translations and Settings → Tafsir download flows. Defaults stay alQuran.cloud (translations) and api.quran.com (tafsir); Bayaan behavior is unchanged when both slots are unset.

Third in the XF-NNN multi-tenancy series, following:
- **RFC-007** ([#235](https://github.com/thebayaan/Bayaan/pull/235), merged) — `Branding` interface + `CatalogProvider`. Same shape applied here.
- **RFC-008** ([#261](https://github.com/thebayaan/Bayaan/pull/261), open) — `branding.listenTabTopComponent`.

## Why two seams in one RFC

`services/translation/TranslationApiService.ts` and `services/tafseer/TafseerApiService.ts` have structurally identical consumer surfaces — both expose `fetchAvailableEditions()` + `fetchFull{X}(editionId, onProgress)` returning `{edition, verses}`, and both stores (`store/translationStore.ts:33`, `store/tafseerStore.ts:39`) use them through the same lifecycle (call → destructure → hand to `dbService.save{X}()`).

Splitting into two RFCs would mean: duplicated reviewer overhead, risk of half-migrated state if one merged and the other stalled, and forks having to maintain a mixed pattern (one seam + one forked file) in the interim. The implementation PRs can still split per-service if that's easier to land — see "Implementation plan" in the RFC.

## Why two parallel interfaces vs. one generic

`Alt 2` in the RFC documents the `ContentEditionProvider<TVerse, TEdition>` generic alternative. The verse shapes differ (`TafseerVerse` carries `groupVerseKey`/`fromAyah`/`toAyah`); the edition shapes differ (`TafseerEdition` carries `authorName`); and unifying the method names would force both store call-sites to change for no real win. Two parallel interfaces compose to the same machinery without the abstraction tax. Open to bikeshedding either way.

## Scope

- **Doc-only.** No code changes in this PR.
- Implementation PR is ~80 lines per service (extract class body → default provider, collapse service file to a resolver, add `Branding` field). Can land as one PR or two.
[...]

PR #266, opened 2026-05-22
feat: branding.listenTabTopComponent slot (RFC-008)

omar-zarka:
## Summary

Implements the **RFC-008** component slot (`docs/rfcs/008-listen-tab-top-slot.md`, merged doc-only via #261).

Adds an optional `listenTabTopComponent` to the `Branding` interface. When a fork sets it, it replaces the Listen-tab top region — currently the hardcoded `<RecitersHero />` in `RecitersView.tsx` — with the fork-provided component. `undefined` (Bayaan's default) keeps `RecitersHero` rendering verbatim.

This retires the last merge-conflict-prone hardcoded JSX line in `RecitersView.tsx`, completing what #260 (`branding.homeRowConfig`) started for the rows below the hero.

## Changes

- `config/branding.d.ts` — `import type {ComponentType}`; new `ListenTabTopComponentProps` interface (empty slot for future extension); new optional `Branding.listenTabTopComponent` field.
- `components/RecitersView.tsx` — conditional render: `listenTabTopComponent` when set, else `RecitersHero`.

~20 lines total. Bayaan behaviour is byte-identical when the slot is unset.

## Test plan

- [x] `tsc --noEmit` clean
- [x] Default path unchanged — `RecitersHero` still renders when `listenTabTopComponent` is undefined

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #263, opened 2026-05-17
feat: branding.catalogVersionEndpoint (RFC-010)

omar-zarka:
## Summary

Adds an optional `branding.catalogVersionEndpoint?: string` field. When set, the app polls this URL on cold-start (and `AppState 'active'`, debounced 5 min) to discover whether the published catalog is newer than what's locally tracked, and triggers a `CatalogProvider` refetch when so. Versioned `url` in the response (CDN-cacheable forever) is preferred over the live mutable catalog URL.

Bayaan default behavior is unchanged: the field is undefined → no polling.

## Motivation

Forks with runtime catalog publishers (Qariah's new ops console at [omar-zarka/qariah-ops-console](https://github.com/omar-zarka/qariah-ops-console), potentially others) need to refresh the catalog without an app rebuild per content change. Today's Bayaan ties catalog updates to `DATA_VERSION` cache-bust on every release.

The cleanest place to add this is the multi-tenant seam established by RFC-007 (#235): a single optional URL each tenant can set.

## What's in this PR

- **`docs/rfcs/010-catalog-version-endpoint.md`** — proposal, alternatives considered (SSE, full-catalog polling, embedded versioning, status quo), backward-compat, security considerations, open questions.
- **`config/branding.d.ts`** — adds the optional `catalogVersionEndpoint?: string` field with full doc on shape + fail-open semantics.
- **`services/catalogVersionPoll.ts`** — small (~95 LOC) implementation: 1.5s timeout, debounced foreground re-poll, AsyncStorage-tracked last-seen version, fail-open on all errors.
- **`services/__tests__/catalogVersionPoll.test.ts`** — 5 tests covering undefined-endpoint no-op, 500 fail-open, parse-error fail-open, refetch-with-versioned-url, no-refetch-when-equal. All pass.

## Shape decision

Followed the RFC-007 idiom (small config field + small service module) — combined doc + code in a single PR since the shape is straightforward. Happy to split into doc-only-first per the RFC-008 (#261) pattern if you'd prefer to discuss shape before code lands.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx jest services/__tests__/catalogVersionPoll.test.ts` → 5/5 pass
- [ ] (Maintainer) review the call-site instructions in the RFC's "Decision" section — no call-site is added in this PR; consumers wire `initCatalogVersionPolling(refetchCatalog)` from their `app/_layout.tsx`.

## Fork consumer

[...]

PR #267, opened 2026-05-22
feat: translation + tafsir provider seams (RFC-009)

omar-zarka:
## Summary

Implements the **RFC-009** content-provider seams (`docs/rfcs/009-content-provider-seams.md`, merged doc-only via #262).

Adds two optional `Branding` slots — `translationProvider` and `tafsirProvider` — so forks can supply their own translation / tafsir sources for the Settings → Translations and Settings → Tafsir download flows. Bayaan's behaviour is unchanged when the slots are unset.

## Changes

- **New interfaces** — `types/TranslationProvider.ts`, `types/TafsirProvider.ts`. Two methods each (`fetchAvailableEditions`, `fetchFull{Translation,Tafseer}`), mirroring the existing service surfaces.
- **Type moves** — `TranslationVerse` → `types/translation.ts`, `TafseerVerse` → `types/tafseer.ts`, so the interfaces + implementations share them. The old locations (`TranslationApiService` / `TafseerApiService`) re-export for back-compat.
- **Default providers** — the current service class bodies are extracted *verbatim* into `AlQuranCloudTranslationProvider` and `QuranComTafsirProvider`.
- **Resolvers** — `TranslationApiService.ts` / `TafseerApiService.ts` become thin `branding.<provider> ?? <default>` resolvers. The exported singleton names + shapes are unchanged → `store/translationStore` and `store/tafseerStore` are untouched.

## Backwards compatibility

None broken. With both slots unset (Bayaan's default), the singletons resolve to the extracted-verbatim default providers — base URLs, response shapes, the RTL `direction` fallback, the paginated by-chapter loop, and the verse-grouping logic all preserved. Neither store changes.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `prettier --check` clean on all touched files
- [x] Default path unchanged — singletons resolve to the verbatim-extracted providers when slots are unset

### DIFF
--- a/config/branding.d.ts
+++ b/config/branding.d.ts
@@ -1,5 +1,7 @@
 import type {ComponentType} from 'react';
 import type {Track} from '@/types/audio';
+import type {TranslationProvider} from '@/types/TranslationProvider';
+import type {TafsirProvider} from '@/types/TafsirProvider';
 
 /** Catalog source config for the active branding. */
 export interface BrandingCatalogConfig {
@@ -176,6 +178,18 @@ export interface Branding {
    * }
    */
   initialPlayerVerseKey?: (track: Track) => string | undefined;
+  /**
+   * RFC-009 — optional translation source for Settings → Translations.
+   * `undefined` keeps Bayaan's default (the alQuran.cloud-backed
+   * `alQuranCloudTranslationProvider`).
+   */
+  translationProvider?: TranslationProvider;
+  /**
+   * RFC-009 — optional tafsir source for Settings → Tafsir. `undefined`
+   * keeps Bayaan's default (the api.quran.com-backed
+   * `quranComTafsirProvider`).
+   */
+  tafsirProvider?: TafsirProvider;
 }
 
 declare const branding: Branding;
--- a/data/availableTafaseer.ts
+++ b/data/availableTafaseer.ts
@@ -1,5 +1,18 @@
 import type {TafseerEdition} from '@/types/tafseer';
 
+/**
+ * Boot-time tafsir editions list, hand-curated and shipped in-bundle.
+ *
+ * RFC-009 — this static list intentionally coexists with the dynamic
+ * `TafsirProvider.fetchAvailableEditions()` runtime path. The static
+ * list is the no-network UI surface used by Settings → Tafsir to render
+ * the picker on first launch and offline; the provider's dynamic
+ * fetcher is for forks that want a server-driven editions catalog
+ * (e.g. surfacing new tafaseer without an app update). Forks may
+ * either: (a) keep this static list and treat `fetchAvailableEditions`
+ * as a future hook, (b) replace it with a `useEffect` that calls
+ * `fetchAvailableEditions`, or (c) merge both.
+ */
 export const AVAILABLE_TAFASEER: TafseerEdition[] = [
   {
     identifier: '169',
--- a/services/tafseer/QuranComTafsirProvider.ts
+++ b/services/tafseer/QuranComTafsirProvider.ts
@@ -1,0 +1,160 @@
+import type {TafseerEdition, TafseerVerse} from '@/types/tafseer';
+import type {TafsirProvider} from '@/types/TafsirProvider';
+
+/**
+ * RFC-009 — Bayaan's default `TafsirProvider`, backed by the
+ * api.quran.com (Quran.com / QF) public API. Extracted verbatim from the
+ * former `TafseerApiService` class body; behaviour is unchanged.
+ */
+
+const API_BASE = 'https://api.quran.com/api/v4';
+
+const RTL_LANGUAGES = new Set([
+  'arabic',
+  'urdu',
+  'kurdish',
+  'persian',
+  'pashto',
+]);
+
+interface QFTafsirResource {
+  id: number;
+  name: string;
+  author_name: string;
+  slug: string;
+  language_name: string;
+}
+
+interface QFTafsirVerse {
+  resource_id: number;
+  verse_key: string;
+  text: string;
+}
+
+interface QFPagination {
+  per_page: number;
+  current_page: number;
+  next_page: number | null;
+  total_pages: number;
+  total_records: number;
+}
+
+interface QFTafsirResponse {
+  tafsirs: QFTafsirVerse[];
+  pagination: QFPagination;
+}
+
+function capitalize(s: string): string {
+  return s.charAt(0).toUpperCase() + s.slice(1).toLowerCase();
+}
+
+class QuranComTafsirProvider implements TafsirProvider {
+  async fetchAvailableEditions(): Promise<TafseerEdition[]> {
+    const res = await fetch(`${API_BASE}/resources/tafsirs`);
+    if (!res.ok) {
+      throw new Error(`API request failed: ${res.status} ${res.statusText}`);
+    }
+    const json = (await res.json()) as {tafsirs: QFTafsirResource[]};
+
+    return json.tafsirs.map(t => {
+      const lang = capitalize(t.language_name);
+      return {
+        identifier: String(t.id),
+        language: lang,
+        name: t.name,
+        englishName: t.name,
+        authorName: t.author_name,
+        format: 'text',
+        type: 'tafsir',
+        direction: RTL_LANGUAGES.has(t.language_name.toLowerCase())
+          ? ('rtl' as const)
+          : ('ltr' as const),
+      };
+    });
+  }
+
+  async fetchFullTafseer(
+    editionId: string,
+    onProgress?: (progress: number) => void,
+  ): Promise<{
+    edition: TafseerEdition;
+    verses: TafseerVerse[];
+  }> {
+    onProgress?.(0);
+
+    // Fetch edition info from available list
+    const editions = await this.fetchAvailableEditions();
+    const edition = editions.find(e => e.identifier === editionId);
+    if (!edition) {
+      throw new Error(`Tafseer edition not found: ${editionId}`);
+    }
+
+    const verses: TafseerVerse[] = [];
+
+    for (let ch = 1; ch <= 114; ch++) {
+      const chapterVerses = await this.fetchChapterTafseer(editionId, ch);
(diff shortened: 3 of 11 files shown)


osmansaeday:
### Code review

osmansaeday (dismissed):
LGTM

osmansaeday:
Conflict with develop after #263 merged. Could you rebase? Approved and CI green otherwise, ready to merge once rebased.

omar-zarka:
Rebased on develop (which now carries #263). The `config/branding.d.ts` conflict with `catalogVersionEndpoint` is resolved — both optional fields kept (`catalogVersionEndpoint` then `translationProvider` then `tafsirProvider`). `tsc --noEmit` clean. Ready to merge.

osmansaeday (dismissed):
LGTM

osmansaeday:
Conflict again after #266 merged (same branding.d.ts issue). Sorry for the back-and-forth. Could you rebase once more? Should be the last one for this batch.

omar-zarka:
Rebased cleanly onto develop (which now carries #266). `config/branding.d.ts` resolved keeping all three optional fields side-by-side (`catalogVersionEndpoint` + `listenTabTopComponent` + `translationProvider` + `tafsirProvider`). `tsc --noEmit` green. Force-pushed to `d48d8ac`. Ready to merge.

osmansaeday (commented):
Clean extraction. Interface in `types/`, providers in `services/<domain>/<vendor>Provider.ts`, branding slots optional so Bayaan needs zero config changes. Backwards-compat story is sound: undefined property access on the CJS export falls through `??` to the bundled defaults.

**HIGH: pre-existing `entry.text.trim()` will throw on `null`** (`services/tafseer/QuranComTafsirProvider.ts` around L130). The `as QFTafsirResponse` cast is unvalidated. If QF ever returns `text: null` for an empty grouped-verse slot, `.trim()` throws TypeError, the chapter fetch aborts, and `fetchFullTafseer` rejects mid-download (the SQLite transaction spans all 115 chapters sequentially, so partial state is possible). One-char fix: `if (entry.text?.trim()) { ... }`. The extraction is the right moment to fix the pre-existing bug.

**MEDIUM: `fetchAvailableEditions` doesn't normalize `direction`** (`types/TranslationProvider.ts` JSDoc). The contract says providers MUST default `direction` when upstream omits it, but only `fetchFullTranslation` does. `AlQuranCloudTranslationProvider.fetchAvailableEditions` skips the normalization. `RemoteTranslationEdition.direction` is typed non-optional but the response is cast unsafely, so a missing field is silently `undefined`. Doesn't bite today (only static `AVAILABLE_TRANSLATIONS` list is used) but it's a trap for whoever wires up a dynamic editions list.

**MEDIUM: `fetchFullTafseer` re-fetches editions internally**. Every download triggers `/resources/tafsirs` for one edition lookup. Pre-existing, but now formalized in the fork contract. Consider passing optional `edition?: TafseerEdition` parameter, or caching.

**LOW**: `types/tafseer.ts:1` comment still says "Tafseer edition from Al Quran Cloud API". Post-RFC-009 this is the shared edition contract used by QF + static list + future custom providers. Make it neutral.

**LOW**: no tests. Verbatim-extraction claim has no automated guard. Worth a follow-up.

**Open questions:**

1. Should the seam include an offline/bundled hook so a fork can skip the download flow entirely (e.g., serving from a binary)?
2. Module-level `??` evaluation means a provider cannot be swapped at runtime. Function-call form (`getTranslationProvider()`) would preserve future flexibility. Intentional?
3. `AVAILABLE_TAFASEER` static list vs `fetchAvailableEditions` dynamic list: both intended to coexist, or is the static one being replaced in a follow-up?

Not approving yet, leaving these for your call. Bayaan-side branding: no changes required (both slots intentionally unset).

omar-zarka:
Thanks for the review — addressed in `ac9c434`.

**HIGH — `entry.text.trim()` null guard**: Fixed. Typed `QFTafsirVerse.text` as `string | null` (reflects the cast-unvalidated reality) and switched to `if (entry.text?.trim())`. When `text` is `null`/`undefined`/whitespace the entry falls through the non-leader branch — joined to the previous leader's group, or dropped if no leader exists yet (same semantics as an empty grouped slot today). Leader's narrowed non-null `text` is carried via a `GroupLeader = QFTafsirVerse & {text: string}` shape so downstream `verses[].text` stays non-null. SQLite transaction integrity preserved — no partial commit risk since the throw was the bug.

**MEDIUM — `fetchAvailableEditions` direction normalization**: Fixed. `AlQuranCloudTranslationProvider.fetchAvailableEditions` now defaults `direction` by language (same `RTL_LANGUAGES` set as `fetchFullTranslation`, fallback `'ltr'`). Contract is now consistently honored across both methods.

**MEDIUM — `fetchFullTafseer` edition param**: Fixed. Interface gains optional third arg `edition?: TafseerEdition`; provider skips the internal `/resources/tafsirs` lookup when a matching edition is passed. `store/tafseerStore.ts` now passes the matching `AVAILABLE_TAFASEER` entry — every boot-time-list download is now one network call shorter. The no-arg path is preserved so any future caller without an edition in scope still works.

**LOW — `types/tafseer.ts:1` comment**: Fixed. Neutralized to describe the shared cross-provider contract (static `AVAILABLE_TAFASEER` + QF default provider + fork overrides). Applied the same treatment to `RemoteTranslationEdition` in `types/translation.ts` for consistency.

**LOW — no tests**: Acknowledged, deferred to a follow-up. The verbatim-extraction guard is a useful safety net; will open a separate PR once the seam shape stabilizes.

---

**Open questions:**

1. **Offline/bundled hook**: Yes long-term, but out of scope for this seam. The intent here was to surface the network-fetch contract first; a `BundledProvider` (serving from a packaged binary, no download flow) is a natural future addition that can ship without breaking this API. Holding it back to keep this PR focused.

2. **Module-level `??` vs function-call form**: Intentional. Both Bayaan and Qariah want compile-time-resolved providers — no per-call indirection cost on the hot tafsir/translation paths. Any runtime-swap need (e.g. user-selectable translation source mid-session) is better served by an explicit higher-level hook than by making the provider itself swappable at runtime. A future fork that wants runtime polymorphism can wrap the export in a function and re-export without touching this contract.

3. **`AVAILABLE_TAFASEER` static vs `fetchAvailableEditions` dynamic**: Intended to coexist — now documented inline on `AVAILABLE_TAFASEER`'s JSDoc. Static is the no-network UI surface (Settings → Tafsir picker on first launch / offline); dynamic is the fork hook for server-driven editions catalogs. Three viable adoption patterns are spelled out: keep static / replace with dynamic / merge both.

`tsc --noEmit` clean; Prettier `--check` clean. Ready for re-review when you have a moment.

osmansaeday (dismissed):
LGTM. All four findings addressed cleanly. The `GroupLeader` intersection type narrowing `text` to `string` is a particularly nice touch, keeps the downstream `verses[].text` non-null without an `as` cast at every reference. The optional `edition?: TafseerEdition` param on `fetchFullTafseer` + tafseerStore passing the static lookup result is exactly the right shape; it preserves the no-arg fallback while removing the `/resources/tafsirs` round-trip on the hot path.

Two minor things for a future iteration (not blocking):

**NIT (pre-existing): `edition.direction` mutation in place** at `AlQuranCloudTranslationProvider.ts:97-99`. Diverges stylistically from the map-style normalization you used in `fetchAvailableEditions`. Pre-existing pattern, not introduced here, but the new map-style version is cleaner and worth converging on when convenient.

**NIT: `GroupLeader` `as string` cast** at `QuranComTafsirProvider.ts:~120`. The enclosing `if (trimmed)` guard confirms non-null but TS can't follow into the spread. A one-line `// trimmed above, non-null guaranteed` comment keeps it readable if strict-mode linting ever flags it.

Open question answers are reasonable across the board: out-of-scope bundled hook is a clean future addition, compile-time `??` resolution matches Bayaan's build-time-selected branding model, and the documented three coexistence patterns for `AVAILABLE_TAFASEER` resolve the static-vs-dynamic ambiguity.

Bayaan-side: `config/branding.js` unchanged, both keys remain optional, `??` fallback selects bundled providers.

Approving + merging.

osmansaeday:
Approval is in but this needs a rebase, develop has moved (RFC-012 / #268 just merged, which also touched `config/branding.d.ts`) and the merge is now CONFLICTING. Could you rebase onto current `develop` and force-push? Should be a quick conflict in `branding.d.ts` where the two RFCs added optional keys to the same `Branding` interface. Will land it once green.

omar-zarka:
Rebased onto current `develop` (post-#268) — same `config/branding.d.ts` conflict pattern, kept RFC-012's `searchFilters` block and appended RFC-009's `translationProvider` + `tafsirProvider` after it. Force-pushed. Ready for merge whenever you are.

osmansaeday (dismissed):
Re-approving after the rebase. Verified `config/branding.d.ts` on the new head: `searchFilters` (RFC-012 from #268) is preserved and the new key(s) interleave cleanly after it. No content drift from the previously-approved diff. CI green. Merging.

osmansaeday:
Re-approved on the previous rebase but #269 (RFC-013) just landed in between and also touched `config/branding.d.ts`, so the merge is back to CONFLICTING. One more rebase needed. Should be the same shape of conflict, `Branding` interface keys interleaving. Will merge as soon as it's green again.

osmansaeday:
Rebased this onto current `develop` (post-#269 merge) for you via maintainer edits. Conflict was purely the `Branding` interface in `config/branding.d.ts`, both PRs add optional keys at the end of the interface, plus the `import type` block. Resolution: kept both `Track` and `TranslationProvider`/`TafsirProvider` imports, and kept all three new keys (`initialPlayerVerseKey` then `translationProvider` / `tafsirProvider`).

No code logic changed, only the conflict resolution. `prettier --check` passes; CI now re-running on `493f325`. Will re-approve and merge once green.

osmansaeday (approved):
Re-approving after the maintainer-edit rebase. branding.d.ts merge resolution is mechanical (interleaved optional keys + both imports). prettier --check passes. Will merge when CI green.

PR #265, opened 2026-05-19
docs(rfc-012): composable Search-tab filter dimensions

omar-zarka:
## Summary

Doc-only RFC proposing composable filter dimensions in the Search tab. Today the Browse-by-X tiles on the Home/Listen tab each route into `BrowseReciters` with a single pre-set filter param, but the filter is invisible to the user once they're there — they can't see what filtered them in, can't relax the filter, and can't combine dimensions.

Two shapes proposed:
- **Option A — declarative config:** `branding.searchFilters?: SearchFilterDimension[]` (shared chip UI, tenant declares which dims to surface).
- **Option B — component slot:** `branding.searchFiltersComponent?` (full tenant control, mirrors RFC-008 pattern).

Either way, the Browse-by-X tiles flip from hidden-filter dead-ends to preset deeplinks into Search with the chip pre-applied.

Motivated by user-feedback on the Qariah fork's 2026-05-19 TestFlight beta walkthrough. Sprint 21 shipped a targeted surahId fix Qariah-side; this RFC proposes the structural fix that benefits Bayaan equally.

## Test plan

- [x] Doc lands in `docs/rfcs/012-composable-search-filters.md`
- [ ] Maintainer feedback on Option A vs B
- [ ] Code-PR follows in a separate cycle after shape clears

Follows the RFC-008 + RFC-009 + RFC-011 doc-only-first cadence.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #268, opened 2026-05-22
feat: branding.searchFilters (RFC-012, v1 — has-photo chip)

omar-zarka:
## Summary

Implements the **v1 shape** of the RFC-012 composable-filter framework (doc-only at #265, accuracy fixes addressed).

Adds an optional `searchFilters` array to the `Branding` interface. When a fork sets it, each id renders a user-editable chip in the Search-tab Browse-Reciters strip. v1 ships the `has-photo` dimension end-to-end as the first array-driven chip — proves the pattern without churning the existing teacher/student chip system. `rewaya` / `has-surah` stay on their current bespoke implementations and migrate in a follow-up; `country` / `translation` / `recitation-style` are flagged in the RFC as future dimensions requiring catalog/type prerequisites and are **not** in this PR.

Bayaan behaviour is byte-identical with `branding.searchFilters` unset.

## Changes

- **`config/branding.d.ts`** — new `SearchFilterDimension` union (exists-today subset only) + optional `Branding.searchFilters?: SearchFilterDimension[]`.
- **`components/browse/BrowseReciters.tsx`** — new `initialHasPhoto?: boolean` prop; togglable "Has photo" chip rendered after the existing chip strip only when `branding.searchFilters` includes `'has-photo'`. Predicate reuses the existing `hasImage` helper (hoisted above the filter chain) so the filter + sort agree on what counts as a photo. `handleClearFilters` + `hasActiveFilters` updated.
- **`app/(tabs)/(a.home)/reciter/browse.tsx`** + **`app/(tabs)/(b.search)/reciter/browse.tsx`** — read the `hasPhoto` URL param; pass through as `initialHasPhoto`. Lets a Browse-by-X tile deeplink in with the chip pre-applied (`?hasPhoto=1`).

## Scope deferrals (acknowledged)

- **Field prerequisites.** The doc PR (#265, fcc8553) was updated to split exists-today dimensions from prerequisites; this code-PR sticks to the exists-today subset. `country` / `translation` / `recitation-style` need `Reciter` (and/or `Rewayat`) fields added before they're implementable, per the review feedback.
- **Migrating `rewaya` + `has-surah` to be array-driven.** Both already work today as bespoke logic (teacher/student chip system + the `surahId` prop). Lifting them into `searchFilters` is a behaviour-preserving refactor; happy to do it in a follow-up once this v1 shape clears, or in this PR if preferred.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `prettier --check` clean on all touched files
- [x] Bayaan default (`branding.searchFilters` undefined) — chip strip + filter pipeline unchanged
- [x] Fork example (`searchFilters: ['has-photo']`) — chip renders, toggling filters the list to reciters with a resolvable photo (catalog `image_url` or bundled headshot)
- [x] URL deeplink (`/reciter/browse?hasPhoto=1`) — chip pre-applied

### DIFF
--- a/app/(tabs)/(a.home)/reciter/browse.tsx
+++ b/app/(tabs)/(a.home)/reciter/browse.tsx
@@ -28,8 +28,8 @@ export default function BrowseScreen() {
   const title = rewayatName
     ? rewayatName
     : surahId
-    ? `Browse Reciters - ${SURAHS[parseInt(surahId, 10) - 1].name}`
-    : 'Browse All';
+      ? `Browse Reciters - ${SURAHS[parseInt(surahId, 10) - 1].name}`
+      : 'Browse All';
 
   // Set native header title on iOS
   useLayoutEffect(() => {
--- a/app/(tabs)/(a.home)/reciter/browse.tsx
+++ b/app/(tabs)/(a.home)/reciter/browse.tsx
@@ -9,12 +9,16 @@ export default function BrowseScreen() {
   const router = useRouter();
   const navigation = useNavigation();
   const {theme} = useTheme();
-  const {surahId, teacher, student, rewayatName} = useLocalSearchParams<{
-    surahId: string;
-    teacher: string;
-    student: string;
-    rewayatName: string;
-  }>();
+  const {surahId, teacher, student, rewayatName, hasPhoto} =
+    useLocalSearchParams<{
+      surahId: string;
+      teacher: string;
+      student: string;
+      rewayatName: string;
+      // RFC-012 — composable Search-tab filter chips. Tiles on the Home
+      // tab can deeplink here with chips pre-applied (e.g. `?hasPhoto=1`).
+      hasPhoto: string;
+    }>();
 
   const handleBack = () => {
     router.back();
@@ -24,8 +28,8 @@ export default function BrowseScreen() {
   const title = rewayatName
     ? rewayatName
     : surahId
-      ? `Browse Reciters - ${SURAHS[parseInt(surahId, 10) - 1].name}`
-      : 'Browse All';
+    ? `Browse Reciters - ${SURAHS[parseInt(surahId, 10) - 1].name}`
+    : 'Browse All';
 
   // Set native header title on iOS
   useLayoutEffect(() => {
@@ -42,6 +46,7 @@ export default function BrowseScreen() {
       title={title}
       initialTeacher={teacher}
       initialStudent={student}
+      initialHasPhoto={hasPhoto === '1' || hasPhoto === 'true'}
     />
   );
 }
--- a/app/(tabs)/(b.search)/reciter/browse.tsx
+++ b/app/(tabs)/(b.search)/reciter/browse.tsx
@@ -7,7 +7,12 @@ import {SURAHS} from '@/data/surahData';
 export default function BrowseScreen() {
   const router = useRouter();
   const {theme} = useTheme();
-  const {surahId} = useLocalSearchParams<{surahId: string}>();
+  // RFC-012 — composable Search-tab filter chips. Tiles can deeplink in
+  // with chips pre-applied (e.g. `?hasPhoto=1`).
+  const {surahId, hasPhoto} = useLocalSearchParams<{
+    surahId: string;
+    hasPhoto: string;
+  }>();
 
   const handleBack = () => {
     router.back();
@@ -24,6 +29,7 @@ export default function BrowseScreen() {
       onBack={handleBack}
       surahId={surahId ? parseInt(surahId, 10) : undefined}
       title={title}
+      initialHasPhoto={hasPhoto === '1' || hasPhoto === 'true'}
     />
   );
 }
--- a/components/browse/BrowseReciters.tsx
+++ b/components/browse/BrowseReciters.tsx
@@ -328,6 +328,7 @@ export default function BrowseReciters({
     surahId,
     advancedFilters,
     hasPhoto,
+    hasPhotoEnabled,
   ]);
 
   const handleFilterModalPress = () => {
@@ -589,6 +590,9 @@ export default function BrowseReciters({
                   </View>
                 ) : (
                   <Pressable
+                    accessibilityRole="togglebutton"
+                    accessibilityLabel={chip.label}
+                    accessibilityState={{selected: chip.isSelected}}
                     style={({pressed}) => [
                       styles.filterChip,
                       chip.isSelected && styles.filterChipActive,
@@ -614,10 +618,12 @@ export default function BrowseReciters({
              * once they're wired through (or live as bespoke chips above). */}
             {hasPhotoEnabled && (
               <Animated.View
-                key="rfc012-has-photo"
                 entering={FadeIn.duration(300)}
                 layout={LinearTransition.duration(300)}>
                 <Pressable
+                  accessibilityRole="togglebutton"
+                  accessibilityLabel="Has photo"
+                  accessibilityState={{selected: hasPhoto}}
                   style={({pressed}) => [
                     styles.filterChip,
                     hasPhoto && styles.filterChipActive,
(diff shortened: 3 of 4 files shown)


omar-zarka:
Rebased on develop (which now carries #263) and applied Osman's #265 review correction: `recitation-style` added to the `SearchFilterDimension` union alongside the other exists-today dimensions (Rewayat.style is live and filtered against in BrowseReciters today). v1's end-to-end wiring stays `has-photo` only; `rewaya` / `has-surah` / `recitation-style` are in the type union so forks can declare them, but they continue routing through their existing bespoke chip systems until later migration sprints.

`tsc --noEmit` clean. The `config/branding.d.ts` rebase resolution keeps both `catalogVersionEndpoint` (#263) and `searchFilters` (this PR) as optional fields.

osmansaeday:
### Code review

Found 1 issue:

1. The nested ternary in `app/(tabs)/(a.home)/reciter/browse.tsx` (`: surahId ? ... : 'Browse All'`) is indented at 4 spaces in this PR but Prettier wants 6 spaces. `npx prettier --check app/\(tabs\)/\(a.home\)/reciter/browse.tsx` reports `[warn] Code style issues found in the above file`; the same check on the pre-PR version passed clean, so this regression is introduced here. The PR description's `prettier --check` claim is also inaccurate. CLAUDE.md says "always run `npx tsc --noEmit` and `npx prettier --write` on changed files. Do not consider work done until both pass clean with zero errors."

https://github.com/thebayaan/Bayaan/blob/9d295735f4cef6c4a1689714e5271a8a1a5eeba1/app/(tabs)/(a.home)/reciter/browse.tsx#L27-L33

omar-zarka:
Fixed in bd1a82f. Root cause: my `npx prettier --check` was resolving to a hoisted Prettier 2.8.8 from a parent `node_modules`, not the repo-pinned 3.8.1. Re-running via `node node_modules/prettier/bin/prettier.cjs --check` reproduced your warning. The Prettier 3 nested-ternary rule wants 6-space indentation on inner branches (was 4-space in Prettier 2). `prettier --write` applied the fix; no other files in the PR diff need re-formatting under Prettier 3.

Standing rule (just added internally): always invoke Prettier via `node node_modules/prettier/bin/prettier.cjs` in Bayaan, not `npx prettier`, until npm's hoisting resolves cleanly.

osmansaeday (commented):
Schema is future-proof and the `hasImage` unification of filter+sort predicate is clean. For Bayaan we're intentionally keeping `searchFilters` unset, so the chip stays invisible on our build. Findings below are on the implementation that should still ship correctly for forks that opt in.

**MEDIUM: missing accessibility props on chip Pressable** (`components/browse/BrowseReciters.tsx`, new chip block). No `accessibilityLabel`, `accessibilityRole`, or `accessibilityState`. Screen-reader users can't discover the toggle semantics. The existing `dynamicFilterChips` Pressable has the same gap (pre-existing), but this PR is the opportunity to set the pattern for all chips:

```tsx
<Pressable
  accessibilityRole="togglebutton"
  accessibilityLabel="Has photo"
  accessibilityState={{selected: hasPhoto}}
  ...
>
```

**LOW: `hasPhotoEnabled` missing from `filteredReciters` useMemo dep array**. Branding is a static CJS singleton today so no stale-closure bug at runtime, but it's a `react-hooks/exhaustive-deps` violation and confuses any future reader who introduces a dynamic branding override.

**LOW: stray `key` prop on sibling `Animated.View` chip**. `key="rfc012-has-photo"` is set on a JSX child of a `ScrollView`, not inside a `.map()`. React ignores keys outside array iterators. The `LinearTransition` / `FadeIn` animations rely on Reanimated's layout-animation system, not React keys, so the animation is unaffected. Remove the prop, or (better) move the chip into the `dynamicFilterChips` data array (which would also be the correct composability path for v2).

**LOW: `branding.d.ts` `SearchFilterDimension` comment cites wrong style string**. Doc says `'molim'` but `data/rewayat-slugs.json` maps that to canonical `"moalim"`. Use the canonical slug to avoid future filter implementers picking the wrong one.

**Open questions:**

1. PR description says Home-tab tiles can deeplink with `?hasPhoto=1` pre-applied, but no Home-tab tile code is modified. Follow-up PR, or should the deeplink wiring be part of this PR to make the chip actually reachable for forks that turn it on?
2. Chip ordering: should `has-photo` render before teacher/student chips, or is the trailing position intended UX?
3. Should `searchFilters: []` be documented as explicit "disable all RFC-012 chips" to distinguish intent from `undefined` ("not yet migrated")?

Not approving. Bayaan-side: `searchFilters` stays unset.

omar-zarka:
Thanks for the review @OKKHALIL3. Addressed all four findings + answered the open questions. Branch rebased onto current `develop` (absorbed #265/#266/#271/#272/#276); pushed as `13f6b26`.

**Per-finding:**

- **MEDIUM — a11y props on chip Pressables.** Added `accessibilityRole="togglebutton"` + `accessibilityLabel` + `accessibilityState={{selected}}` to the new `has-photo` chip Pressable. Also applied the same pattern to the existing `dynamicFilterChips` Pressable (teacher/student/all chips) so the whole view is consistent — you flagged the pre-existing gap as the right opportunity to set the pattern, agreed.
- **LOW — `hasPhotoEnabled` missing from `filteredReciters` useMemo deps.** Added. Agreed it's a defensive read for the day a future fork wires a dynamic branding override.
- **LOW — stray `key=\"rfc012-has-photo\"` on sibling `Animated.View`.** Removed. Confirmed locally that the `LinearTransition` / `FadeIn` animations are unaffected (Reanimated's layout-animation system tracks via its own id, not React keys). Holding off on moving the chip into the `dynamicFilterChips` data array — that's the v2 composability path you noted, planning to do it as part of the bespoke-chip migration RFC rather than complicating this PR.
- **LOW — `'molim'` → `'moalim'` canonical slug.** Fixed in the `SearchFilterDimension` JSDoc; also corrected the inline trailing-comment on the `'recitation-style'` variant in the same union to cite `'murattal'|'mojawwad'|'moalim'` per `data/rewayat-slugs.json`.

**Open questions:**

1. **Home-tile deeplink wiring.** Follow-up PR, intentionally — scope of #268 is the seam + the `has-photo` chip end-to-end inside `BrowseReciters`. The Home-tile changes are conceptually a separate slice (touches different surface, different reviewer mental model), so will land alongside the `rewaya` / `has-surah` / `recitation-style` v2 migration where the same deeplink-tile pattern will need to be wired for all chips at once.
2. **Chip ordering — trailing position intended.** Yes, deliberate. New RFC-012 dimensions added by forks should appear after Bayaan's existing bespoke chips (teacher/student) so a Bayaan user opening a forked build sees the familiar chips first and the fork's additions after. Documented this convention explicitly in the JSDoc on `Branding.searchFilters` in this push.
3. **`searchFilters: []` semantics.** Documented the tri-state in the JSDoc on `Branding.searchFilters`:
   - `undefined` (default) — \"not migrated\"; today's bespoke chip set unchanged. Bayaan ships this.
   - `[]` — \"explicitly disable all RFC-012 chips\". Observably the same as `undefined` today, but the intent diverges in v2 when bespoke chips themselves move under this seam.
   - non-empty array — opt in to the listed dimensions.

Ready for another look when you have a cycle.

osmansaeday (approved):
LGTM. All four findings addressed, plus the a11y triad extended to the pre-existing `dynamicFilterChips` Pressable (good call). The bonus work is solid too: hoisting `hasImage` so filter + sort share one definition, threading `hasPhoto` through `handleClearAllFilters` and `hasActiveFilters`, and the `hasPhotoEnabled && initialHasPhoto` guard on the deeplink param so a stale `?hasPhoto=1` can't force-activate the filter on a build where the chip is disabled.

Verified `chip.label` and `chip.isSelected` exist on every non-separator chip in `dynamicFilterChips`, so the `accessibilityLabel={chip.label}` / `accessibilityState={{selected: chip.isSelected}}` pattern is correct.

Tri-state JSDoc on `searchFilters` is the right call. Bayaan-side: leaving `searchFilters` unset (chip stays invisible in Bayaan), per the v1-rollout decision.

Merging.

PR #264, opened 2026-05-17
RFC-011: Continue-Reading hero history size

omar-zarka:
## Summary

Doc-only RFC proposing a small numeric Branding field — `continueReadingHistorySize?: number`, defaulting to `1` — that widens the Surahs-tab CONTINUE READING hero from a single card into a horizontal carousel of the last N stopping points. Bayaan's behavior is unchanged when the field is omitted.

## Motivation

Multi-session readers (users juggling daily-review surah / memorization surah / study surah simultaneously) lose context when the single `lastReadPage` MMKV slot gets overwritten. Currently the only workaround is to fork both `services/mushaf/MushafSessionStore.ts` and `components/hero/ContinueReadingHero.tsx` — which is what Qariah is doing today, paying a merge cost on every Bayaan touch to either file.

The seam is small: ~80 lines of Bayaan-side diff if accepted; preserves the existing `getLastReadPage()` API via a shim; adds a new `getRecentPages()` for the carousel consumer. Persistence migration of legacy values is transparent.

## Shape (RFC-008 / RFC-010 precedent)

Doc-only first, code PR follows once the maintainer signs off on the shape. The numeric-config-on-Branding pattern matches RFC-010 (`catalogVersionEndpoint`) more than RFC-008 (`listenTabTopComponent`) — the alternative of "make this a component slot" is considered and rejected in the RFC.

## Test plan

- [ ] Review proposed `Branding.continueReadingHistorySize?: number` field shape
- [ ] Confirm the proposed `MushafSessionStore` shim (legacy `getLastReadPage()` → `getRecentPages()[0]?.page ?? null`) is acceptable
- [ ] Confirm the `{page, openedAt}` entry shape for the new `RecentPage` type
- [ ] Open questions in section 5 — clamping strictness + `openedAt` seeding on legacy migration

Per the existing Qariah ↔ Bayaan collaboration cadence (see RFC-007 / RFC-008 / RFC-010 precedent): doc-only PR first; code PR follows once shape is approved.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #269, opened 2026-05-23
feat: branding.initialPlayerVerseKey (RFC-013, combined doc + code)

omar-zarka:
## Summary

Adds an optional `branding.initialPlayerVerseKey?: (track: Track) => string | undefined` hook letting forks tell QuranView (PlayerSheet ayah list) where to anchor the scroll on cold-mount and on surah changes. `undefined` (default) keeps the existing scroll-to-top-of-surah behavior **byte-identical**.

Motivating use case: range-restricted recitations — audio tracks covering only a subset of a surah (e.g. Qariah's 326 partial tracks across 54 reciters, where audio for Al-Baqarah may play ayahs 197-202 only). Without this seam, the ayah list scrolls to ayah 1 while the audio plays ayahs 197-202 — a real, actively-shipped UX defect.

Per the RFC-008 + RFC-011 pattern, this is doc + code together because the shape is small and self-contained (one hook, opt-in, default-undefined → no Bayaan behavior change).

## Changes

- **`config/branding.d.ts`** — adds `initialPlayerVerseKey?: (track: Track) => string | undefined`. Imports `Track` from `@/types/audio`.
- **`components/player/v2/PlayerContent/QuranView/index.tsx`** —
  - Subscribes to the current track via granular `usePlayerStore` selector.
  - Memoizes `initialScrollIndex` from the hook's return value + `verses.findIndex`.
  - Passes `initialScrollIndex` to FlashList (handles cold-mount natively during initial layout pass).
  - Rewrites the surah-change effect to defer `scrollToIndex` two `requestAnimationFrame`s when an initial verse_key is in play — needed because FlashList v2 silently no-ops `scrollToIndex` calls made before the new \`data\` has finished laying out (no `onScrollToIndexFailed` escape hatch like FlatList).
  - Default branch (no hook → no initial verse_key → `scrollToOffset({offset: 0})`) preserved verbatim.
  - Drive-by: wraps `const verses = enhancedVersesBySurah[currentSurah] ?? []` in `useMemo` so the new useMemo/useEffect deps are stable. Also clears a pre-existing react-hooks/exhaustive-deps warning.
- **`docs/rfcs/013-initial-player-verse-key.md`** — full RFC text describing design, alternatives considered, open questions, risks.

## Default Bayaan behavior

Identical to before. `branding.initialPlayerVerseKey` is undefined in Bayaan's `config/branding.js`, so `initialScrollIndex` resolves to `undefined`, FlashList ignores it, and the surah-change effect falls into the `else { scrollToOffset({offset: 0}) }` branch. The only observable diff from this PR alone is that `verses` is now memoized — same value, same behavior, just a stable reference.

## Test plan

- [x] `npx jest` — 116/116 passing
- [x] `node node_modules/prettier/bin/prettier.cjs --check` on changed files — clean (Prettier 3.8.1)
- [x] `npx eslint` on changed files — zero new errors/warnings introduced
- [x] Verified the Qariah fork's equivalent code path on iPhone 17 Pro sim: tapping a Hajar Boussaq partial recitation (Al-Baqarah Warsh, ayahs 197-202) opens the PlayerSheet with the ayah list anchored on 2:197 instead of 2:1 (commit 4aa0dc3 on the Qariah fork uses the equivalent shape inline; this PR lifts it to a `branding` seam so the fork can retire the divergence)
- [ ] (Reviewer) Bayaan-side smoke: open the PlayerSheet for any full-surah recitation → expect scroll-to-top of surah on each surah change (unchanged behavior)
- [ ] (Reviewer) Bayaan-side smoke: tap active-ayah autoscroll lock + unlock → expect autoscroll-to-current-verse on highlight tick (unchanged behavior; the new defer only fires when the hook returns a verse_key)

## Related

- RFC-007 (multi-tenant config seam, merged)
- RFC-008 (listen-tab top component slot, merged)
- RFC-010 (catalog version endpoint, merged)
- RFC-011 (continue-reading history size, in flight #264)

## Follow-up (not in this PR)

- Migrate the existing active-ayah autoscroll effect (L218-230) to the 2-rAF defer pattern as a small reliability tweak. Currently relies on the implicit fact that the autoscroll fires "long after" track-load, which holds today but breaks if a fork wires the autoscroll on the first audio tick. Punted to a follow-up chore PR.

### DIFF
--- a/components/player/v2/PlayerContent/QuranView/index.tsx
+++ b/components/player/v2/PlayerContent/QuranView/index.tsx
@@ -1,5 +1,11 @@
 import React, {useCallback, useRef, useEffect, useState, useMemo} from 'react';
-import {View, StyleSheet, Pressable, useWindowDimensions} from 'react-native';
+import {
+  View,
+  StyleSheet,
+  Pressable,
+  useWindowDimensions,
+  type LayoutChangeEvent,
+} from 'react-native';
 import {moderateScale, verticalScale} from '@/utils/scale';
 import {useResponsive} from '@/hooks/useResponsive';
 import {Ionicons} from '@expo/vector-icons';
@@ -166,6 +172,22 @@ export const QuranView: React.FC<QuranViewProps> = ({
   // Counter to force re-render when enhanced verses are rebuilt (async)
   const [, setRebuildCounter] = useState(0);
 
+  // Measured ListHeaderComponent height (SurahDivider + BasmalaHeader). Used
+  // as a negative `viewOffset` on imperative scrollToIndex calls so the
+  // target verse lands fully below the header rather than partially behind
+  // it. FlashList v2's `initialScrollIndex` + `scrollToIndex` compute their
+  // target offset from item layouts only; without this compensation, the
+  // 5% viewPosition lands inside the header band. Stored in a ref so a
+  // measurement change doesn't itself trigger re-render of the surah-change
+  // effect — the ref is read at scroll time, by which point onLayout has
+  // run. Initialized to 0 (no compensation) → first paint may briefly
+  // under-shoot; the 2-rAF deferred scrollToIndex below corrects it once
+  // the header has measured.
+  const headerHeightRef = useRef(0);
+  const handleHeaderLayout = useCallback((e: LayoutChangeEvent) => {
+    headerHeightRef.current = e.nativeEvent.layout.height;
+  }, []);
+
   // DK Skia rendering: derive font family and fontMgr from mushafRenderer
   const isDK =
     (mushafRenderer === 'dk_v1' ||
@@ -234,6 +256,22 @@ export const QuranView: React.FC<QuranViewProps> = ({
   // mount, `initialScrollIndex` on FlashList handles it natively, so this
   // effect's branch is mainly for subsequent in-app surah switches where
   // the list stays mounted.
+  //
+  // `viewOffset: -headerHeight` compensates for ListHeaderComponent
+  // (SurahDivider + BasmalaHeader). Without this, FlashList v2 computes
+  // the target scroll offset from item layouts only and the verse lands
+  // partially behind the header. `viewPosition: 0.05` then places the
+  // verse 5% down from the post-header viewport top.
+  //
+  // Multi-surah double-fire note: in single-surah Bayaan tracks, a track
+  // change always changes currentSurah, so this effect fires exactly once.
+  // If multi-surah tracks ship later (e.g. Juz-spanning recitations), the
+  // dep array can fire twice — once from `currentSurah` advancing within
+  // the same track, once from `initialScrollIndex` resolving on a later
+  // render. Future implementer adding multi-surah support should add a
+  // ref-based guard like `if (lastFiredForSurahRef.current === currentSurah
+  // && lastFiredForIndexRef.current === initialScrollIndex) return;`
+  // before the imperative scroll. Out of scope for RFC-013 v1.
   useEffect(() => {
     if (!listRef.current) return;
     setIsLocked(true);
@@ -245,6 +283,7 @@ export const QuranView: React.FC<QuranViewProps> = ({
               index: initialScrollIndex,
               animated: false,
               viewPosition: 0.05,
+              viewOffset: -headerHeightRef.current,
             });
           } catch {
             listRef.current?.scrollToOffset({offset: 0, animated: false});
@@ -264,7 +303,12 @@ export const QuranView: React.FC<QuranViewProps> = ({
     if (index === -1) return;
 
     try {
-      listRef.current.scrollToIndex({index, animated: true, viewPosition: 0.3});
+      listRef.current.scrollToIndex({
+        index,
+        animated: true,
+        viewPosition: 0.3,
+        viewOffset: -headerHeightRef.current,
+      });
     } catch {
       // Index may be out of range during recycling — ignore
     }
@@ -350,20 +394,22 @@ export const QuranView: React.FC<QuranViewProps> = ({
         keyExtractor={keyExtractor}
         initialScrollIndex={initialScrollIndex}
         ListHeaderComponent={
-          <QuranListHeader
-            surahNumber={currentSurah}
-            showBismillah={!!surah?.bismillah_pre}
-            width={contentWidth}
-            textColor={readingColors.textSecondary}
-            nameColor={readingColors.text}
-            showTajweed={showTajweed}
-            fontMgr={fontMgr}
-            dkFontFamily={dkFontFamily}
-            indexedTajweedData={indexedTajweedData}
-            arabicTextWeight={arabicTextWeight}
-            showAllahNameHighlight={showAllahNameHighlight}
-            allahNameHighlightColor={allahNameHighlightColor}
-          />
+          <View onLayout={handleHeaderLayout}>
+            <QuranListHeader
+              surahNumber={currentSurah}
+              showBismillah={!!surah?.bismillah_pre}
+              width={contentWidth}
+              textColor={readingColors.textSecondary}
+              nameColor={readingColors.text}
+              showTajweed={showTajweed}
+              fontMgr={fontMgr}
+              dkFontFamily={dkFontFamily}
+              indexedTajweedData={indexedTajweedData}
+              arabicTextWeight={arabicTextWeight}
+              showAllahNameHighlight={showAllahNameHighlight}
+              allahNameHighlightColor={allahNameHighlightColor}
+            />
+          </View>
         }
         contentContainerStyle={{
           paddingTop: effectivePaddingTop,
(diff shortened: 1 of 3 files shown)


osmansaeday (commented):
Solid RFC. Seam is minimal, opt-in, Bayaan's default path is byte-identical to pre-PR. The selector granularity (`s => s.queue.tracks[s.queue.currentIndex]`) avoids re-renders on position/duration ticks. No Bayaan branding changes needed: `undefined` is the correct representation of our "scroll to top of surah" default.

**MEDIUM: `initialScrollIndex` on FlashList does not account for `ListHeaderComponent` height**. `QuranView` uses `ListHeaderComponent` (renders `SurahDivider` + `BasmalaHeader`). FlashList v2's `applyInitialScrollIndex` computes the scroll offset from item layouts directly, without adding the rendered header height. So when `initialScrollIndex > 0`, the list lands at approximately the right item but the target verse renders partially behind the divider/Basmala instead of at `viewPosition: 0.05`. The 2-rAF deferred `scrollToIndex` path has the same offset semantics. This is the first place in the codebase where `initialScrollIndex` and `ListHeaderComponent` coexist on a FlashList. Fix options: pass `viewOffset` equal to the measured header height, or use a `viewPosition` that compensates.

Not a crash and not a Bayaan regression (Bayaan doesn't set this key), but the feature won't look polished on any surah whose Basmala header takes meaningful pixel space.

**LOW: double-fire on multi-surah Juz tracks**. The useEffect dep array includes `[currentSurah, initialScrollIndex, setIsLocked]`. For Bayaan's single-surah tracks this is fine. If multi-surah tracks ship later, the scroll may fire twice (once from `currentSurah` change, once from `initialScrollIndex` resolving asynchronously on a subsequent render). RFC's Open Question 1 acknowledges the surah-tab case conceptually but doesn't note the double-fire possibility. Worth an inline comment or a guard.

**LOW: JSDoc example uses `+track.surahId` without the guard from the Migration plan**. A reader copying just the JSDoc example could produce a `NaN`-keyed `getSurahMetadata` lookup. Tighten the JSDoc guard to match the Migration example, or replace `+track.surahId` with `parseInt(track.surahId, 10)`.

**LOW: RFC status is "Draft"** but the doc ships with code. Combined-doc-and-code RFC should land as `Accepted` or `Implemented`. Check RFC-008/011 precedent.

**Open questions:**

1. Has the Qariah fork smoke-tested `initialScrollIndex` specifically for surahs with a long Basmala header (where the offset miss is visible)?
2. FlashList v2 consumes `initialScrollIndex` only on construction. When the PlayerSheet fully unmounts and remounts (app background then a different track is now playing), does the new mount pick up the new `initialScrollIndex` cleanly in your test matrix?

Not approving, holding for the header-offset finding before merge.

omar-zarka:
Thanks for the thorough review. Addressed in `221e104`. Also rebased on current `develop` (clean — three of my own PRs absorbed in the interim, no conflicts on the RFC-013 surface).

## What changed

**MEDIUM — `viewOffset` for `ListHeaderComponent`.** Wrapped the `ListHeaderComponent` in a `<View onLayout={...}>` that writes the measured height into a `headerHeightRef`, and pass `viewOffset: -headerHeightRef.current` to **all three** `scrollToIndex` call sites:

1. The 2-rAF deferred surah-change scroll (the RFC-013 mount path).
2. The active-ayah autoscroll effect (L260-271, pre-existing — same offset bug, fixed in passing).
3. The recenter button (L383-401, pre-existing — same).

Went with `onLayout` measurement rather than a constant because the header height isn't actually constant — `BasmalaHeader` derives its height from Skia paragraph metrics (font + width-scaled glyph rendering) and `SurahDivider` from typeface metrics scaled to `contentWidth`. A constant would either over- or under-shoot on tablet split-view, larger Dynamic Type, or when `parentContentWidth` is passed by the tablet layout. The ref is read at scroll time (post-onLayout), so even though it initializes to `0`, the 2-rAF defer guarantees the measurement has landed before the imperative scroll fires. I considered using the existing `computeDividerTotalHeight(contentWidth)` helper but that only covers the divider — Basmala height is private to the cached Skia paragraph build, so `onLayout` is the simpler unified path.

Note `initialScrollIndex` is untouched (FlashList consumes it on construction with no `viewOffset` parameter), but the 2-rAF deferred `scrollToIndex` fires on the same mount and refines the final position with the compensation applied.

**LOW — multi-surah double-fire comment.** Added a block comment on the surah-change `useEffect` explicitly documenting the risk and what the future fix shape looks like (`lastFiredForSurahRef` / `lastFiredForIndexRef` guard). For single-surah Bayaan tracks the current dep array fires exactly once per track change, so no behavior change today.

**LOW — JSDoc guard.** Tightened the `@example` in `branding.d.ts` to match the Migration plan example exactly: `parseInt(track.surahId, 10)` + `Number.isNaN` guard + explicit `meta.is_full` / `meta.range` checks. Also added the same `Number.isNaN` guard to the Migration plan code block so the two examples stay in lockstep — readers can copy either one safely.

**LOW — RFC status.** Changed `Draft` → `Accepted` and reformatted the header to the canonical markdown-table style used by RFC-008 / RFC-010 / RFC-011, with the combined-doc-and-code note moved to a blockquote underneath.

## Answers to open questions

1. **Long-Basmala smoke test on Qariah.** Honest answer: Greta verified the partial-ayah seek functionally on the Pixel 7 emulator earlier today — the verse now opens at the correct ayah (Sprint-23 carry-forward, pre-RFC-013 absorption). She did not specifically eyeball viewport position relative to the Basmala band. The `viewOffset` fix in this PR will land in Qariah's next beta (S24) and I'll verify visually on a long-Basmala surah (Al-Baqarah:255 partial is a good probe — Basmala renders + ayah is well below it). Will report back here once confirmed.

2. **Unmount + remount with a new `initialScrollIndex`.** FlashList v2's `RecyclerViewManager.getInitialScrollIndex()` reads the prop on construction; a full unmount + remount with a different prop value should land at the new index per the v2 source (`useRecyclerViewController.js#applyInitialScrollIndex` keys on `initialScrollCompletedRef` which resets with the component). I have **not** specifically tested the "background app + different track on resume" path — the 2-rAF deferred `scrollToIndex` was added precisely for the case where `initialScrollIndex` is stale relative to a subsequent data change. If the PlayerSheet retains its mount on background+resume (which I believe it does — RN AppState transitions don't tear down the React tree), the deferred path covers it. Adding "background-then-different-track" to Qariah's S24 test matrix.

---

Diff stat: `3 files changed, 80 insertions(+), 23 deletions(-)`. `npx tsc --noEmit` clean. `prettier --check` clean.

osmansaeday (dismissed):
LGTM. All four findings addressed.

The dynamic `onLayout`-measured `headerHeightRef` + `viewOffset: -headerHeightRef.current` at all three scroll call sites was the right call over a hard-coded constant, Skia paragraph metrics aren't stable across Dynamic Type, tablet split-view, and varying `parentContentWidth`. The 2-rAF defer ensures `onLayout` has fired before the imperative scroll, so the ref won't be read as `0` in the deferred path. The double-fire inline comment with named future `lastFiredForSurahRef` / `lastFiredForIndexRef` guard shape is precise enough to be actionable when multi-surah tracks ship.

Two new nits surfaced on the re-review (not blocking):

**NIT: FlashList JSX block has no `estimatedItemSize` override for the anchored item.** FlashList v2's `applyInitialScrollIndex` estimates pixel offset as `initialScrollIndex * estimatedItemSize` before real cell heights are known. Verse items have variable height (WBW, translation, transliteration toggles), so the pre-layout estimate will be off, potentially causing a visible re-jump once measure runs. The 2-rAF `scrollToIndex` corrects the final position immediately, so this is cosmetic (transient scroll flicker), not a functional defect. Worth noting if you ever see a flash on the Qariah anchored-verse path.

**NIT: RFC text inconsistency.** Section "Open questions" item 2 says `useMemo([trackReciterId, trackRewayatId, currentSurah, verses])` but the shipped implementation uses `[currentTrack, verses]`. `currentTrack` is the whole track reference, which churns more often than just the three fields the hook cares about (the selector `s => s.queue.tracks[s.queue.currentIndex]` returns a new reference on most queue state changes). Either tighten the selector to extract just the fields, or update the RFC text to match the shipped dep array.

Bayaan-side: no config change. `branding.initialPlayerVerseKey` undefined preserves "scroll to top of surah" default.

Approving + merging.

osmansaeday:
Approval is in but this needs a rebase, develop has moved (RFC-012 / #268 just merged, which also touched `config/branding.d.ts`) and the merge is now CONFLICTING. Could you rebase onto current `develop` and force-push? Should be a quick conflict in `branding.d.ts` where the two RFCs added optional keys to the same `Branding` interface. Will land it once green.

omar-zarka:
Rebased onto current `develop` — quick conflict resolution in `config/branding.d.ts` (kept RFC-012's `searchFilters` block then appended RFC-013's `initialPlayerVerseKey` after it). Force-pushed. Ready for merge whenever you are.

osmansaeday (approved):
Re-approving after the rebase. Verified `config/branding.d.ts` on the new head: `searchFilters` (RFC-012 from #268) is preserved and the new key(s) interleave cleanly after it. No content drift from the previously-approved diff. CI green. Merging.

PR #273, opened 2026-05-23
fix: race in mushaf font loading (useFonts vs preload)

omar-zarka:
## Problem

Mushaf-rendering Skia components currently run a `useFonts(...)` call alongside reading the preloaded `mushafPreloadService.fontMgr`:

```ts
const hookFontMgr = useFonts({ /* ...DK + SurahNameV4 + ... */ });
const fontMgr = mushafPreloadService.fontMgr || hookFontMgr;
```

The comment in each call site says the `useFonts` call is a "fallback" — but it always runs (can't conditionally call hooks). On first render, when many `<SurahDivider>` instances mount simultaneously (e.g. opening the Player's QuranView, or scrolling the continuous mushaf), the parallel `useFonts` calls race against each other and against the preload service's own `Skia.Typeface.MakeFreeTypeFaceFromData` calls, producing intermittent:

```
Couldn't create typeface for SurahNameV4
```

errors. These show up in Sentry but don't fail the user-visible render (the preloaded fontMgr usually wins by the time anything's actually drawn). Still noise we shouldn't ship.

## Fix

Subscribe to the preload service via `useSyncExternalStore` instead.

- `MushafPreloadService.subscribe(listener)` — minimal new method; notifies on init completion.
- `hooks/useMushafFontMgr.ts` — new hook that returns `mushafPreloadService.fontMgr` synchronously and re-renders when the preload finishes.
- Three call sites switch from `useFonts` + preload-OR-fallback to the new hook:
  - `components/mushaf/skia/SkiaPage.tsx`
  - `components/mushaf/skia/ContinuousMushafView.tsx`
  - `components/player/v2/PlayerContent/QuranView/SurahDivider.tsx`

The consumer pattern collapses from:

```ts
const hookFontMgr = useFonts({ /* five entries */ });
const fontMgr = mushafPreloadService.fontMgr || hookFontMgr;
```

to:

```ts
const fontMgr = useMushafFontMgr();
```

AppInitializer (priority 5) awaits `mushafPreloadService.initialize()` before splash hides, so in practice `fontMgr` is already non-null on first render. The subscription only matters for the brief window where a Skia component mounts before preload completion — e.g. a modal that snaps open during AppInitializer's run.

## Scope

`Header.tsx` (player chrome) keeps its own `useFonts` for `SurahNameQCF` — it's outside the Mushaf-render path and its font set doesn't overlap the contention case. `MushafPageIcon.tsx` and `ReadingThemeContent.tsx` already read only from the preload service (no `useFonts` fallback), so they're already correct, just don't re-render on late preload completion; that's a separate (lower-priority) follow-up.

## Validation

Shipped in the Qariah fork on 2026-05-17 (Sprint 17, story S17.11). No `Couldn't create typeface` reports in Qariah's Sentry since. Local `tsc --noEmit` clean.

Suggest local sim verification on Bayaan before merge — open the Mushaf tab, scroll a few pages, open the player's QuranView, confirm no Skia typeface warnings in Metro logs.

## Diff stat

```
5 files changed, 61 insertions(+), 45 deletions(-)
```

### DIFF
--- a/components/mushaf/skia/ContinuousMushafView.tsx
+++ b/components/mushaf/skia/ContinuousMushafView.tsx
@@ -11,11 +11,11 @@ import {FlashList, type FlashListRef} from '@shopify/flash-list';
 import {
   Canvas,
   Skia,
-  useFonts,
   type SkFont,
   type SkParagraph,
   type SkTypefaceFontProvider,
 } from '@shopify/react-native-skia';
+import {useMushafFontMgr} from '@/hooks/useMushafFontMgr';
 import {Gesture, GestureDetector} from 'react-native-gesture-handler';
 import {runOnJS} from 'react-native-worklets';
 import * as Haptics from 'expo-haptics';
@@ -736,21 +736,9 @@ const ContinuousMushafView = forwardRef<
     const render = useMemo(() => buildRenderConstants(metrics), [metrics]);
     const {lineWidth} = render;
 
-    // Font loading (fallback — prefer preloaded fontMgr)
-    const hookFontMgr = useFonts({
-      DigitalKhattV1: [
-        require('@/data/mushaf/legacy/DigitalKhattQuranicV1.otf'),
-      ],
-      DigitalKhattV2: [
-        require('@/data/mushaf/digitalkhatt/DigitalKhattFont.otf'),
-      ],
-      DigitalKhattIndoPak: [
-        require('@/data/mushaf/indopak/DigitalKhattIndoPak.otf'),
-      ],
-      QuranCommon: [require('@/data/mushaf/quran-common.ttf')],
-      SurahNameV4: [require('@/data/mushaf/surah-name-v4.ttf')],
-    });
-    const fontMgr = mushafPreloadService.fontMgr || hookFontMgr;
+    // Subscribe to preloaded fontMgr; no useFonts fallback that races
+    // at first-mount.
+    const fontMgr = useMushafFontMgr();
 
     // Surah header fonts (computed once from quranCommon typeface)
     const surahHeaderFonts = useMemo(() => {
--- a/components/mushaf/skia/ContinuousMushafView.tsx
+++ b/components/mushaf/skia/ContinuousMushafView.tsx
@@ -740,7 +740,11 @@ const ContinuousMushafView = forwardRef<
     // at first-mount.
     const fontMgr = useMushafFontMgr();
 
-    // Surah header fonts (computed once from quranCommon typeface)
+    // Surah header fonts (computed once from quranCommon typeface).
+    // `fontMgr` dep mirrors the SkiaPage memo: when the preload completes
+    // after this view mounted, the subscription re-renders but `lineWidth`
+    // hasn't changed, so without `fontMgr` in deps the memo would stay on
+    // its cached null-divider snapshot.
     const surahHeaderFonts = useMemo(() => {
       const qcTypeface = mushafPreloadService.quranCommonTypeface;
       if (!qcTypeface) return {dividerFont: null, nameFontSize: 0};
@@ -753,7 +757,7 @@ const ContinuousMushafView = forwardRef<
         dividerFont: Skia.Font(qcTypeface, scaledSize),
         nameFontSize: scaledSize * 0.4,
       };
-    }, [lineWidth]);
+    }, [lineWidth, fontMgr]);
 
     // Settings subscriptions
     const showTajweed = useMushafSettingsStore(s => s.showTajweed);
@@ -776,8 +780,8 @@ const ContinuousMushafView = forwardRef<
       (mushafRenderer === 'dk_indopak'
         ? 'DigitalKhattIndoPak'
         : mushafRenderer === 'dk_v1'
-        ? 'DigitalKhattV1'
-        : 'DigitalKhattV2');
+          ? 'DigitalKhattV1'
+          : 'DigitalKhattV2');
     const allahNameHighlightColor = useMemo(
       () =>
         getAllahNameHighlightColorHex(
--- a/components/mushaf/skia/SkiaPage.tsx
+++ b/components/mushaf/skia/SkiaPage.tsx
@@ -1,11 +1,7 @@
 import React, {useMemo, useState, useEffect, useRef, useCallback} from 'react';
 import {View, Platform} from 'react-native';
-import {
-  Canvas,
-  Skia,
-  useFonts,
-  type SkParagraph,
-} from '@shopify/react-native-skia';
+import {Canvas, Skia, type SkParagraph} from '@shopify/react-native-skia';
+import {useMushafFontMgr} from '@/hooks/useMushafFontMgr';
 import {Gesture, GestureDetector} from 'react-native-gesture-handler';
 import {runOnJS} from 'react-native-worklets';
 import * as Haptics from 'expo-haptics';
@@ -106,21 +102,10 @@ const SkiaPage: React.FC<SkiaPageProps> = ({
   const PAGE_PADDING_TOP = propPaddingTop ?? DEFAULT_PAGE_PADDING_TOP;
 
   const {theme} = useTheme();
-  // Keep useFonts hook as fallback (can't conditionally call hooks).
-  // Prefer preloaded fontMgr from MushafPreloadService; ready synchronously
-  // on first render since AppInitializer runs before Mushaf tab mounts.
-  const hookFontMgr = useFonts({
-    DigitalKhattV1: [require('@/data/mushaf/legacy/DigitalKhattQuranicV1.otf')],
-    DigitalKhattV2: [
-      require('@/data/mushaf/digitalkhatt/DigitalKhattFont.otf'),
-    ],
-    DigitalKhattIndoPak: [
-      require('@/data/mushaf/indopak/DigitalKhattIndoPak.otf'),
-    ],
-    QuranCommon: [require('@/data/mushaf/quran-common.ttf')],
-    SurahNameV4: [require('@/data/mushaf/surah-name-v4.ttf')],
-  });
-  const fontMgr = mushafPreloadService.fontMgr || hookFontMgr;
+  // Subscribe to the preloaded fontMgr instead of running a parallel
+  // useFonts() fallback that races at first-mount and surfaces as
+  // "Couldn't create typeface for SurahNameV4" Sentry exceptions.
+  const fontMgr = useMushafFontMgr();
 
   // Calculate rendering dimensions (hoisted above surahHeaderFonts so lineWidth is available)
   const scale = CONTENT_WIDTH / PAGE_WIDTH;
--- a/components/mushaf/skia/SkiaPage.tsx
+++ b/components/mushaf/skia/SkiaPage.tsx
@@ -112,7 +112,13 @@ const SkiaPage: React.FC<SkiaPageProps> = ({
   const margin = MARGIN * scale;
   const lineWidth = CONTENT_WIDTH - 2 * margin;
 
-  // Create scaled SkFont objects for surah header rendering (Skia Text path)
+  // Create scaled SkFont objects for surah header rendering (Skia Text path).
+  // `fontMgr` is in the dep array even though we read `quranCommonTypeface`
+  // directly from the singleton: when the subscription fires (preload
+  // completes after this component mounted), `lineWidth` hasn't changed
+  // and the memo would otherwise return its cached null-divider result.
+  // Re-running on the fontMgr transition rebuilds the divider with the
+  // now-available typeface.
   const surahHeaderFonts = useMemo(() => {
     const qcTypeface = mushafPreloadService.quranCommonTypeface;
     if (!qcTypeface) return {dividerFont: null, nameFontSize: 0};
@@ -128,7 +134,7 @@ const SkiaPage: React.FC<SkiaPageProps> = ({
       dividerFont: Skia.Font(qcTypeface, scaledSize),
       nameFontSize: scaledSize * 0.4,
     };
-  }, [lineWidth]);
+  }, [lineWidth, fontMgr]);
 
   const showTajweed = useMushafSettingsStore(s => s.showTajweed);
   const showThemes = useMushafSettingsStore(s => s.showThemes);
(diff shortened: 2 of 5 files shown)


osmansaeday (changes requested):
Direction is right. Serializing font registration via `useSyncExternalStore` is the correct architectural fix for parallel typeface-registration contention (the root cause of the "Couldn't create typeface for SurahNameV4" Sentry exceptions). Notify ordering (fontMgr set before `_notify` fires) is sound. But the implementation has one HIGH severity regression that needs to be addressed before merge, plus two follow-up gaps.

**HIGH: removed fallback exposes a silent-failure path that permanently blanks the Mushaf**

`services/mushaf/MushafPreloadService.ts` `loadSkiaFonts` (L99 to L125) wraps everything in a `try/catch` that swallows all errors with `console.warn`. When the catch fires, `_fontMgr` stays `null`. `_doInit` then continues to `this._initialized = true; this._notify()`, notifying all subscribers that loading is "complete" while `fontMgr` is permanently `null`.

Before this PR, all three call sites used `mushafPreloadService.fontMgr || hookFontMgr`. If `fontMgr` was null from a silent failure, `useFonts` (the hookFontMgr) would eventually resolve and provide a working fallback. After the PR, `useMushafFontMgr()` returns `null` forever. All three components hit `if (!fontMgr) return null`, the Mushaf renders blank, no recovery path.

The same permanent-blank outcome occurs if `_doInit` itself rejects before reaching `_notify` (e.g., `digitalKhattDataService.initialize()` rejects). AppInitializer registers the service as `critical: false`, so the splash hides regardless.

Pick one of:

- **(a)** Rethrow from `loadSkiaFonts` so the outer init can observe the failure.
- **(b)** Add a boolean error state on the service and fire `_notify` from a `finally` block so subscribers can render an error UI.
- **(c)** Keep a lightweight `useFonts` fallback gated on `fontMgr === null` after an initialization timeout.

**MEDIUM: `surahHeaderFonts` useMemo in SkiaPage will not recompute on the subscription-triggered re-render**

`components/mushaf/skia/SkiaPage.tsx` L131 to L146. The `useMemo` reads `mushafPreloadService.quranCommonTypeface` directly and lists `[lineWidth]` as its only dependency. When `useMushafFontMgr` fires a re-render because preload completed after mount, `lineWidth` hasn't changed, so `useMemo` returns its cached `{dividerFont: null, nameFontSize: 0}`. Page renders with a valid `fontMgr` but missing surah header fonts until something else changes `lineWidth` (e.g., orientation change). The PR comment specifically names "fast-mounted modals, dev reloads" as the target case for the subscription, which is exactly the window this miss surfaces in. Fix: add `fontMgr` to the dep array.

**MEDIUM: 14 components still read `mushafPreloadService.fontMgr` directly without subscribing**

This is a partial migration. The following sites read `fontMgr` at render time with no subscription, so they will permanently see `null` if they mount before preload completes:

- `components/MushafPageIcon.tsx:53`
- `components/ReadingThemeContent.tsx:193`
- `components/MushafSettingsContent.tsx:619`
- `components/sheets/VerseShareSheet.tsx:71`
- `components/sheets/SimilarVersesSheet.tsx:90`
- `components/sheets/verse-actions/ShareContent.tsx:81`
- `components/sheets/verse-actions/WBWContent.tsx:59`
- `components/share/SkiaVersePreview.tsx:57`
- `components/mushaf/qcf/QCFPage.tsx:137`
- `components/mushaf/reading/ReadingPageView.tsx:129`
- `components/mushaf/reading/ContinuousListView.tsx:247`
- `components/player/v2/PlayerContent/QuranView/index.tsx:180`

Sheets in particular can open before the Mushaf tab is visited (deep link into a verse action), so these are realistic race windows. Not a regression introduced by this PR, but the PR leaves the system in a partially-migrated state that reads as "fully fixed" from the title. Either document a migration plan in `useMushafFontMgr.ts` JSDoc or roll the hook out to all of them in this PR.

**LOW**: when `_doInit` rejects, `_initPromise` holds a rejected promise. A second `initialize()` call returns the same rejection. No retry path, no way for subscribers to distinguish "loading" from "failed". Subscriber-model gap, now observable in more places than before.

**Open questions:**

1. Was the `loadSkiaFonts` try/catch error swallowing intentional? If so, what's the acceptable UX for cold-start font failure: error state, or fall back to system fonts?
2. Module-singleton `mushafPreloadService` + Fast Refresh: does `_initialized = true` persist across JS reloads without `_notify` firing for newly-mounted components? Dev-only, but worth validating given the hook's docstring mentions fast-mounted modals.

Requesting changes on the HIGH bug. The two MEDIUMs would be best addressed here too; I'll defer the 14 other components to a follow-up if you'd prefer to keep the PR scoped tight.

omar-zarka:
Thanks for the review @OKKHALIL3. New commit `a624739` addresses everything. Per-finding:

## HIGH — silent-blank regression

Picked a hybrid of your (b) + (c). The architectural intent is **invisible fallback** — users never see a blank Mushaf even when preload fails cold-start.

**`MushafPreloadService` changes:**
- New `'idle' | 'loading' | 'ready' | 'failed'` state machine with a `loadError` getter.
- `_doInit` now wraps the whole pipeline in `try/catch/finally`. The `finally` block always transitions to a terminal state (`ready` or `failed`) and always calls `_notify()`, so a hydration/DK/diff rejection no longer leaves subscribers stuck on a `loading` snapshot forever.
- `loadSkiaFonts` is now two-stage: the original parallel load attempts every family; on partial failure (any null typeface or any thrown error) a new `loadSkiaFontsSequentialFallback` retries with serialized `await` between each typeface registration. Sequential ordering defuses the parallel typeface-creation contention that was the root cause of the original `Couldn't create typeface for SurahNameV4` Sentry exceptions, so most failures are recovered transparently inside the service.

**`useMushafFontMgr` changes:**
- Subscribes to both `fontMgr` and `loadError` via `useSyncExternalStore`.
- In the catastrophic case (preload failed AND its sequential retry also produced nothing), the hook falls through to a per-component `useFonts(...)` covering the Mushaf-critical font families (`DigitalKhattV1/V2/IndoPak`, `QuranCommon`, `SurahNameV4`, `SurahNameQCF`). That fallback re-introduces the original parallel-typeface race, but it is bounded to the truly-catastrophic case where the service's own retry has already been exhausted. The alternative — a permanently-blank Mushaf — is the user-facing regression we must not ship.
- Happy path is unchanged: `loadError === false` means `useFonts(...)` is still invoked (Rules of Hooks — must run unconditionally) but its result is discarded.

Net effect: cold-start font failure produces a working Mushaf via the service's sequential retry (most cases) or the hook's `useFonts` fallback (catastrophic case). No error state is surfaced to the user. Sentry still captures the original failure so observability is preserved.

## MEDIUM — `surahHeaderFonts` useMemo stale on subscription re-render

Added `fontMgr` to the dep array in both `SkiaPage.tsx` and `ContinuousMushafView.tsx` (same anti-pattern in both). Comment on each site explains why we depend on a value the memo body doesn't reference: it's a proxy for the subscription's "preload now complete" transition.

## MEDIUM — 14 partial-migration sites

Spinning out as a follow-up to keep #273 tight: #277 migrates 13 of the 14 (12 from your list + one bonus, `SimilarVersesContent.tsx`, which has the same anti-pattern). The 14th (`main.tsx:853`) is a callback-time read, not a render-time read, so it doesn't need the hook. Documented the migration plan inline in `useMushafFontMgr.ts` JSDoc pointing at #277. Stack: #277 is branched off `fix/mushaf-font-loading-race`, so it depends on this PR.

## LOW — rejected `_initPromise` cached forever

`initialize()` now clears `_initPromise` and resets `_state` to `idle` on a prior `failed`, so a second `initialize()` call drives a fresh retry instead of returning the cached rejection. End-to-end retry semantics now work.

## Open questions

**(1) Was the try/catch error swallowing intentional?** No — it was defensive coding that didn't anticipate the subscription model would expose it as a permanent-blank failure mode. Fixed via the loadError + invisible-fallback architecture above. UX intent: cold-start font failure produces a working Mushaf via service-side sequential retry or per-component `useFonts` fallback; no error state shown to user; Sentry still captures the failure.

**(2) Fast Refresh + module singleton.** This is a real dev-only gap. With `_state = 'ready'` persisted across JS reloads, newly-mounted components subscribed via `useSyncExternalStore` will get the `ready` snapshot on subscribe (via the `getSnapshot` return), so they won't hang. But the subscription's `_notify` only fires on state changes, so they won't trigger a fresh render either — they mount with the stored state and render. Acceptable dev-only behavior. If we ever see a dev-only repro of stuck-loading after Fast Refresh, the fix is a module-scope `onHot` or `onModuleReplaced` hook that re-fires `_notify`; not adding it speculatively.

Validated:
- `npx tsc --noEmit` clean both repos
- `node node_modules/prettier/bin/prettier.cjs --write` clean on all 4 changed files in #273 and all 13 changed files in #277

osmansaeday (changes requested):
Significant progress on the architecture, the state machine (`'idle'` → `'loading'` → `'ready'` / `'failed'`), the `finally` block guaranteeing terminal-state notify, the `loadError` field, and the `useFonts` fallback in `useMushafFontMgr` gated on `loadError && !preloadFontMgr` are all the right shapes. The original notification-with-null-fontMgr bug is gone on the primary path.

One remaining HIGH gap and three secondary findings before merge.

---

**HIGH: `loadSkiaFontsSequentialFallback` unconditionally assigns `this._fontMgr = fontMgr` even when zero typefaces loaded**

`services/mushaf/MushafPreloadService.ts`, the new sequential-retry method (last ~60 lines of the diff).

The fallback runs after the primary `loadSkiaFonts` has set `anyFailed = true` and (correctly) withheld `_fontMgr` assignment. The fallback iterates each font family individually. When every `Skia.Typeface.MakeFreeTypeFaceFromData` call returns null (catastrophic device-level Skia failure), the inner `continue` skips every `registerFont` call, the outer `try` succeeds (no exception thrown), and at the bottom:

```ts
this._fontMgr = fontMgr;  // ← empty SkTypefaceFontProvider, but still a non-null object
```

In `_doInit`, the null-guard `if (!this._fontMgr) { failed = true; }` evaluates to `false` (non-null object), so `_state` becomes `'ready'`, `loadError` stays `false`, and `useMushafFontMgr`'s fallback branch (`if (loadError && !preloadFontMgr)`) never activates. The Mushaf renders blank with the empty provider, with no recovery path.

This is structurally the same failure mode the original HIGH finding described, just on the catastrophic-Skia path instead of the swallowed-exception path. The primary `loadSkiaFonts` is correctly hardened (`anyFailed` tracking + conditional assignment); the sequential fallback needs the same guard:

```ts
// Track per-typeface failures across the fallback loop.
let anyCriticalFailed = false;
// ... inside the loop, on a critical-font miss:
anyCriticalFailed = true;

// At the end:
if (!anyCriticalFailed) {
  this._fontMgr = fontMgr;
} else {
  console.warn('[MushafPreload] Sequential fallback failed; useFonts fallback will activate');
  // Leave _fontMgr null so _doInit's null-guard fires.
}
```

Or, more conservatively, check that at least the critical Digital Khatt typeface is present before assigning.

---

**MEDIUM: two `useSyncExternalStore` calls register independent listeners per hook instance**

`hooks/useMushafFontMgr.ts` lines 57-68. `preloadFontMgr` and `loadError` each call `mushafPreloadService.subscribe(listener)` with separate closures, so one `_notify()` iterates two listeners and schedules two React re-renders per state transition. React batches these in concurrent mode, but the codebase isn't bridgeless-universally yet, under legacy rendering this is a double-render on every preload state change.

Combine into a single store slice with a stable snapshot comparator, or derive `isReady` from a single field since `loadError` and `fontMgr` always flip together at `_notify`:

```ts
const isReady = useSyncExternalStore(
  listener => mushafPreloadService.subscribe(listener),
  () => mushafPreloadService.state === 'ready' || mushafPreloadService.state === 'failed',
);
const preloadFontMgr = isReady ? mushafPreloadService.fontMgr : null;
const loadError = isReady && mushafPreloadService.state === 'failed';
```

---

**MEDIUM: `FALLBACK_FONT_SOURCES` comment misleadingly claims sync with `FONT_ASSETS`**

`hooks/useMushafFontMgr.ts` ~L54: comment says "Kept in sync with the `FONT_ASSETS` map in `MushafPreloadService`." `FONT_ASSETS` has 8 entries (including `ManropeSemiBold`, `ManropeRegular`); `FALLBACK_FONT_SOURCES` has 6 (Manrope intentionally excluded since Manrope is a UI font, not Mushaf-critical).

The omission is correct, but the comment will mislead the next maintainer into adding Manrope unnecessarily. One-line fix: "Mushaf-critical subset of `FONT_ASSETS`, UI fonts (Manrope*) are intentionally excluded."

---

**LOW: retry path doesn't reset partial state**

`MushafPreloadService.ts`, `initialize()` retry clears `_initPromise` and resets `_state` to `'idle'`, but leaves whatever partial values the first run wrote (`_quranCommonTypeface` and `_surahNameTypeface` might be set from a partially-successful first run). `_doInit` then creates a fresh `fontMgr` and re-registers families, overwriting in place. Probably harmless, but the invariant should be explicit, reset `_fontMgr`, `_quranCommonTypeface`, `_surahNameTypeface` to `null` at the top of the retry block.

---

**LOW: PR description scope drift**

Description says "5 files changed, 61 insertions(+), 45 deletions(-)", that's the original scope, the current diff is +277/-92. Worth updating for future bisect/blame readability.

---

Two unresolved characterizations on the 14-component migration: you mentioned `MushafPageIcon.tsx` and `ReadingThemeContent.tsx` are "already correct" because they don't use `useFonts`. That's true for the `useFonts`-contention fix but NOT for the subscribe-on-preload-completion fix. Components that read `mushafPreloadService.fontMgr` directly at render time will still see `null` if they mount before preload completes. AppInitializer's splash gate makes this lower-severity now, but the PR description should not describe them as fully addressed. A follow-up issue with the 14 names linked from `useMushafFontMgr.ts` header would make the punt explicit and trackable.

Requesting changes on the HIGH bug. The two MEDIUMs would ideally land here too. Happy to re-review on push.

omar-zarka:
Superseded by #277 — the migration PR includes #273's full service rewrite + hook inline, and Osman's #277 review explicitly notes 'merging #277 obviates #273, and #273 should be closed (or marked superseded) when this lands.' Closing in favor of #277. The HIGH bug carried forward from this review (sequential-fallback empty-fontMgr publish) + the two MEDIUMs + LOW are being addressed against #277.

PR #274, opened 2026-05-23
fix(browse-grid): clear mini-player overlay with bottomInset

omar-zarka:
## Summary

- `BrowseGrid`'s `gridContainer.paddingBottom` was a fixed `moderateScale(80)`. When the floating mini-player + tab bar is showing, the last grid row sits behind that overlay — partially hidden and sometimes untappable.
- This adds an optional `bottomInset` prop (default 0) that grows the bottom padding to `max(80, bottomInset + 16)`. Backward-compatible — callers that don't pass it get the existing behavior.
- `BrowseReciters` reads `useBottomInset()` (already in the codebase, returns 0 when nothing overlays) and threads it through.

## Test plan

- [x] `tsc --noEmit` clean
- [ ] Sim/emulator verification deferred to maintainer. Repro: open Browse All, start any track, scroll to the bottom of the grid — last row should clear the mini-player.

### DIFF
--- a/components/browse/BrowseGrid.tsx
+++ b/components/browse/BrowseGrid.tsx
@@ -6,6 +6,7 @@ import {
   ActivityIndicator,
 } from 'react-native';
 import {moderateScale} from 'react-native-size-matters';
+import {moderateScale as moderateScaleCapped} from '@/utils/scale';
 import {LegendList} from '@legendapp/list';
 import {Reciter} from '@/data/reciterData';
 import {Theme} from '@/utils/themeUtils';
@@ -21,7 +22,8 @@ interface BrowseGridProps {
   // Optional bottom inset to clear the floating mini-player + tab bar.
   // Callers compute via `useBottomInset` so the last grid row stays
   // tappable when the mini-player is visible. Defaults to 0 for callers
-  // that don't need it.
+  // that don't need it; the original `moderateScale(80)` floor still
+  // applies when the inset is small or unset.
   bottomInset?: number;
 }
 
@@ -32,7 +34,7 @@ function createStyles(_theme: Theme, bottomInset = 0) {
     },
     gridContainer: {
       paddingHorizontal: moderateScale(10),
-      paddingBottom: Math.max(moderateScale(80), bottomInset + moderateScale(16)),
+      paddingBottom: Math.max(moderateScaleCapped(80), bottomInset),
       paddingTop: moderateScale(8),
     },
     row: {
@@ -160,6 +162,7 @@ const BrowseGrid = React.memo(
           renderItem={renderRow}
           keyExtractor={keyExtractor}
           contentContainerStyle={styles.gridContainer}
+          scrollIndicatorInsets={{bottom: bottomInset}}
           estimatedItemSize={itemDimensions.height}
           recycleItems
           drawDistance={2000}
--- a/components/browse/BrowseGrid.tsx
+++ b/components/browse/BrowseGrid.tsx
@@ -18,16 +18,21 @@ interface BrowseGridProps {
   keyboardShouldPersistTaps?: 'always' | 'handled' | 'never';
   onScrollBeginDrag?: () => void;
   getRewayatIdForReciter?: (reciter: Reciter) => string | undefined;
+  // Optional bottom inset to clear the floating mini-player + tab bar.
+  // Callers compute via `useBottomInset` so the last grid row stays
+  // tappable when the mini-player is visible. Defaults to 0 for callers
+  // that don't need it.
+  bottomInset?: number;
 }
 
-function createStyles(_theme: Theme) {
+function createStyles(_theme: Theme, bottomInset = 0) {
   return StyleSheet.create({
     container: {
       flex: 1,
     },
     gridContainer: {
       paddingHorizontal: moderateScale(10),
-      paddingBottom: moderateScale(80),
+      paddingBottom: Math.max(moderateScale(80), bottomInset + moderateScale(16)),
       paddingTop: moderateScale(8),
     },
     row: {
@@ -61,6 +66,7 @@ const BrowseGrid = React.memo(
     theme,
     onScrollBeginDrag,
     getRewayatIdForReciter,
+    bottomInset = 0,
   }: BrowseGridProps) => {
     const {width: windowWidth} = useWindowDimensions();
     const [isLoading] = useState(false);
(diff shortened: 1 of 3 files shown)


osmansaeday (commented):
Fix is directionally correct (last grid row was hidden behind the floating mini-player on modern iPhones). Three issues survive review.

**MEDIUM: phantom ~15pt bottom gap on modern iPhones when no track is playing** (`components/browse/BrowseGrid.tsx` around L30). `useBottomInset()` on iOS always includes `insets.bottom` (34pt on iPhone X/14/15) regardless of whether a track is loaded. So with no player active on iPhone 14 Pro, the hook returns `tab_bar + safe_area = ~83pt`. The new expression evaluates to `Math.max(~85, 83 + ~17) = ~100pt` instead of the previous ~85pt. Every notch-or-dynamic-island iPhone now has a permanent ~15pt extra whitespace at the bottom of the browse grid in the idle state. PR description says "returns 0 when nothing overlays" but the hook unconditionally includes `tab_bar + insets.bottom`; it only conditionally adds the player height. Either drop the `+16` buffer (`useBottomInset` already gives comfortable clearance) or restructure the hook to omit safe-area in the idle case.

**LOW: `keyboardShouldPersistTaps` prop is a silent no-op** (L18, L159). The `BrowseGridProps` interface declares it and `BrowseReciters` passes `"handled"`, but L159 hardcodes `"handled"` on the `LegendList` instead of reading from props. Pre-existing dead prop, not introduced by this PR, but the memo comparator change at ~L178 doesn't include `keyboardShouldPersistTaps` either, so a parent passing a different value would also be silently swallowed. Good opportunity to wire it through or remove it from the interface.

**LOW: sibling `CollectionGrid` still has the same bug**. `components/collection/CollectionGrid.tsx:89` has the identical hardcoded `paddingBottom: moderateScale(80)` that this PR just fixed in `BrowseGrid`. Same overlay behavior, same last-row obstruction in the Collections tab. The PR is not wrong to scope narrowly, but the sibling screen exhibits the bug.

**NIT**: `scrollIndicatorInsets` is not set to match the new dynamic `paddingBottom`. On iOS glass builds the vertical scroll bar can extend visually over the mini-player. Sibling screens (`RecitersView`, `SurahsView`) use `contentInsetAdjustmentBehavior="automatic"`; `LegendList` here gets neither.

**NIT**: `moderateScale` imported from `react-native-size-matters` (uncapped) rather than `@/utils/scale` (tablet-capped). On iPad the `+16` buffer inflates freely. Inconsistent with the project's stated scale policy.

**Open questions:**

1. PR description says `useBottomInset` returns 0 when nothing overlays, but it returns `tab_bar + insets.bottom`. Is the idle ~15pt gap acceptable, or should the `+16` buffer drop?
2. Should `CollectionGrid` get the same fix in a follow-up on this branch?
3. Remove `keyboardShouldPersistTaps` from the interface or wire it through?

Not approving, holding for the idle-gap and CollectionGrid sibling at minimum.

omar-zarka:
Thanks for the careful review — pushed `af39375` rebased onto current `develop`. Per-finding response below.

**MEDIUM — phantom ~15pt idle gap.** Fixed by dropping the `+moderateScale(16)` buffer entirely. You're right that `useBottomInset` unconditionally includes `tab_bar + insets.bottom`; on modern iPhones that's already ~83pt of clearance in the idle state, so the extra `+16` was inflating the bottom rail visibly. The expression is now:

```ts
paddingBottom: Math.max(moderateScaleCapped(80), bottomInset)
```

The 80pt floor is preserved (matches the original pre-PR baseline; protects callers that don't pass a `bottomInset` or pass 0). And the PR description's "returns 0 when nothing overlays" claim was just wrong — corrected.

**LOW — sibling `CollectionGrid`.** Fixed in this PR — `components/collection/CollectionGrid.tsx` gains the identical `bottomInset?: number` prop, the same `Math.max(moderateScaleCapped(80), bottomInset)` paddingBottom, matching `scrollIndicatorInsets`, and the memo comparator now considers it. No call sites currently pass it (the prop defaults to 0 → floor wins, same as the previous hardcoded value), but the shape is ready for the Collections tab caller to wire `useBottomInset` whenever convenient.

**NIT — `scrollIndicatorInsets`.** Added `scrollIndicatorInsets={{bottom: bottomInset}}` to both `LegendList`s. `contentInsetAdjustmentBehavior="automatic"` would be the alternative but it interacts with `useNativeHeader` differently than the existing `marginTop` math in `BrowseReciters`, so explicit insets felt safer here.

**NIT — `moderateScale` import.** The only `moderateScale` call on the touched paddingBottom is now switched to `@/utils/scale`'s tablet-capped variant (imported as `moderateScaleCapped` to keep the existing `react-native-size-matters` imports untouched for the rest of the file). The 80pt floor will no longer inflate on iPad. Per scoping discipline I left the other call sites in these two files on `react-native-size-matters` — happy to file a separate cleanup if you'd prefer the whole files migrated.

**LOW — `keyboardShouldPersistTaps` dead prop.** Not touched in this PR — agree it's worth fixing but the choice (wire-through vs remove) depends on whether `BrowseReciters` / any other call site actually relies on overriding the default. Happy to file a separate cleanup PR; let me know if you'd prefer wire-through (interface kept) or removal from the props.

**Open question answers:**
1. Idle ~15pt gap: not acceptable. Dropped the `+16` per your recommendation. The new expression is `Math.max(moderateScaleCapped(80), bottomInset)` — `useBottomInset` already provides comfortable clearance in both idle (~83pt) and player-active (~147pt on iPhone) states.
2. `CollectionGrid`: yes, fixed in this same PR with the identical-shape change.
3. `keyboardShouldPersistTaps`: separate follow-up. Will open a chore PR if you want me to take it.

Verified `tsc --noEmit` clean. No new visual changes besides the bottom-rail tightening.

osmansaeday (approved):
LGTM. All five findings addressed cleanly.

Dropping the `+16` buffer was the right call. `useBottomInset` on a modern iPhone in idle state returns ~83pt (49pt tab bar + 34pt safe-area), which exceeds the 80pt floor, so `bottomInset` wins via `Math.max` and there's no phantom gap. With a track loaded the hook returns 147pt, clearing the mini-player.

`CollectionGrid` fix is structurally identical and inert today (no callers pass `bottomInset` yet), so it's a forward-compatible no-op until the Collections screen wires `useBottomInset`. Worth a follow-up to do that wiring, but not blocking this PR.

The `moderateScaleCapped` alias from `@/utils/scale` makes intent clear at the padding call site without disturbing the other `react-native-size-matters` usages in the file. `scrollIndicatorInsets={{bottom: bottomInset}}` cleans up the iOS glass-build scroll-bar issue.

Deferring `keyboardShouldPersistTaps` cleanup to a separate chore PR is reasonable given the wire-through vs. removal decision depends on intent at the call sites.

Approving + merging.

PR #275, opened 2026-05-23
chore(types): use ListRenderItemInfo for renderItem callbacks

omar-zarka:
## Summary

Several `renderItem` callbacks across sheets / search / collection were typed inline as `({item}: {item: T})`, which omits the `index` + `separators` fields that the list libraries actually pass. Replace with the exported types so the full render-item contract is visible at the callsite.

- `react-native` FlatList → `ListRenderItemInfo<T>`
- `@shopify/flash-list` FlashList → `ListRenderItemInfo<T>` from FlashList (aliased to `FlashListRenderItemInfo` where it coexists with RN's in the same file)
- `@legendapp/list` LegendList → `LegendListRenderItemProps<T>`

Mechanical cleanup, no runtime change.

Files touched:
- `components/CollectionView.tsx`
- `components/collection/CollectionGrid.tsx`
- `components/mushaf/MushafSearchView.tsx` (3 callbacks)
- `components/search/SearchView.tsx` (2 callbacks)
- `components/sheets/FavoriteRecitersSheet.tsx`

## Test plan

- [x] `tsc --noEmit` clean
- [ ] No runtime behavior change — type-only edit

### DIFF
--- a/components/CollectionView.tsx
+++ b/components/CollectionView.tsx
@@ -1,5 +1,10 @@
 import React, {useCallback} from 'react';
-import {View, FlatList, StyleSheet} from 'react-native';
+import {
+  View,
+  FlatList,
+  StyleSheet,
+  type ListRenderItemInfo,
+} from 'react-native';
 import {useTheme} from '@/hooks/useTheme';
 import {moderateScale, verticalScale} from 'react-native-size-matters';
 import {CircularReciterCard} from './cards/CircularReciterCard';
@@ -45,7 +50,7 @@ export default function CollectionView({
   });
 
   const renderItem = useCallback(
-    ({item}: {item: Reciter | TrackItem}) => {
+    ({item}: ListRenderItemInfo<Reciter | TrackItem>) => {
       if ('surahId' in item) {
         const reciter = favoriteReciters.find(r => r.id === item.reciterId);
         return (
--- a/components/collection/CollectionGrid.tsx
+++ b/components/collection/CollectionGrid.tsx
@@ -2,7 +2,7 @@ import React, {useMemo, useCallback} from 'react';
 import {View, StyleSheet, useWindowDimensions} from 'react-native';
 import {moderateScale} from 'react-native-size-matters';
 import {moderateScale as moderateScaleCapped} from '@/utils/scale';
-import {LegendList} from '@legendapp/list';
+import {LegendList, type LegendListRenderItemProps} from '@legendapp/list';
 import {Theme} from '@/utils/themeUtils';
 import {PlaylistCard} from '@/components/cards/PlaylistCard';
 import {CircularReciterCard} from '@/components/cards/CircularReciterCard';
@@ -255,7 +255,7 @@ export const CollectionGrid = React.memo(
 
     // Render a row of items
     const renderRow = useCallback(
-      ({item}: {item: CollectionItem[]}) => (
+      ({item}: LegendListRenderItemProps<CollectionItem[]>) => (
         <View style={styles.row}>
           {item.map(collectionItem => renderCard(collectionItem))}
           {/* Add empty placeholders for incomplete rows */}
--- a/components/mushaf/MushafSearchView.tsx
+++ b/components/mushaf/MushafSearchView.tsx
@@ -10,8 +10,12 @@ import {
   Pressable,
   Animated as RNAnimated,
   BackHandler,
+  type ListRenderItemInfo,
 } from 'react-native';
-import {FlashList, type ListRenderItemInfo} from '@shopify/flash-list';
+import {
+  FlashList,
+  type ListRenderItemInfo as FlashListRenderItemInfo,
+} from '@shopify/flash-list';
 import {moderateScale as ms} from 'react-native-size-matters';
 import {useSafeAreaInsets} from 'react-native-safe-area-context';
 import {useTheme} from '@/hooks/useTheme';
@@ -698,7 +702,7 @@ const MushafSearchView: React.FC<MushafSearchViewProps> = ({
   // Browse mode renderers
   // ──────────────────────────────────────────────────────────
   const renderBrowseItem = useCallback(
-    ({item}: ListRenderItemInfo<BrowseItem>) => {
+    ({item}: FlashListRenderItemInfo<BrowseItem>) => {
       if (item.type === 'juz-header') {
         return (
           <View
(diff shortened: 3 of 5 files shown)


osmansaeday (commented):
Four of the five changes are correct. `CollectionGrid` (LegendList) and the FlashList-vs-RN alias trick in `MushafSearchView` are well executed: aliasing the FlashList import to `FlashListRenderItemInfo` while importing RN's `ListRenderItemInfo` separately removes any future ambiguity. No `any` introduced, no `as` casts, explicit generics throughout.

**MEDIUM: `MushafSearchView.tsx` `renderHistoryItem` is dead code, now annotated with a misleading type**. The function is defined at ~L820 and the PR annotates its parameter as `ListRenderItemInfo<SearchHistoryItem>` from `react-native`. But `renderHistoryItem` is never passed to any list component. The search history renders via an inline `searchHistory.map(...)` at ~L960. The PR's annotation now codifies an incorrect assumption: it implies the function is a `FlatList` callback when it's actually never called that way at all. Either delete the function (it's dead) or revert just this annotation (leave the inline `{item: SearchHistoryItem}` shape since it's not a render callback).

**NIT**: `LegendListRenderItemProps<T>` has `extraData: any`. Not introduced by this PR but now formalized in `CollectionGrid`. Worth a typed `extraData` in a follow-up.

**Question**: is `renderHistoryItem` intentionally dead, or was the inline `.map()` introduced later and the function forgotten? If dead, delete it.

Not approving, holding for the dead-code question.

omar-zarka:
Addressed in 68f4489 (force-pushed after rebase onto current `develop`).

**Fixed (MEDIUM):** Deleted the dead `renderHistoryItem` function. To your question — yes, it's dead. `git grep -n 'renderHistoryItem'` confirms zero call sites; the inline `searchHistory.map(...)` at ~L960 is the only consumer of `HistoryRow`. Most likely the inline `.map()` was introduced later and the `useCallback` was forgotten. The `ListRenderItemInfo` import stays (still used by `renderSearchResult` at ~L813).

**NIT acknowledged:** Agreed — `LegendListRenderItemProps<T>` carrying `extraData: any` is a pre-existing rough edge in `@legendapp/list`'s types, just now formalized by `CollectionGrid`'s explicit generic. Worth a typed `extraData` follow-up if/when we have a concrete shape; happy to file a separate small PR for it.

Diff: 1 file, -12 lines. `tsc --noEmit` clean.

osmansaeday (commented):
Dead `renderHistoryItem` cleanly deleted. The `ListRenderItemInfo` from `react-native` import replacing the old FlashList import is correctly consumed elsewhere.

One small annotation nit worth fixing in the same PR before merge:

**NIT: `renderSearchResult` in `MushafSearchView.tsx` is annotated with the wrong `ListRenderItemInfo`**. The file now has two imports:

- `FlashListRenderItemInfo` (from `@shopify/flash-list`), used by `renderBrowseItem` which is passed to `<FlashList>` at L894. Correct.
- `ListRenderItemInfo` (from `react-native`), used by `renderSearchResult` at L810. But `renderSearchResult` is also passed to a `<FlashList>` at L1008, not a `FlatList`.

TypeScript won't catch this because the two `ListRenderItemInfo` types are structurally identical at the fields actually used (`{item, index, separators}`), but the annotation is semantically wrong and will mislead the next reader who assumes the RN import means FlatList. One-line annotation change.

Other angles all clean: `CollectionView` (FlatList), `CollectionGrid` (LegendList), `SearchView` (FlatList), `FavoriteRecitersSheet` (FlatList) all have correct alignment between list component and `ListRenderItemInfo` source.

Bayaan-side: nothing.

Holding approval on this one annotation fix.

omar-zarka:
Want to double-check before I push the swap — I might be looking at the wrong thing. When I read the file at the branch HEAD (`68f4489`) I'm seeing `renderSearchResult` consumed at L998 by a `<FlatList>` rather than a `<FlashList>`:

```
$ grep -nE 'FlashList|FlatList|renderSearchResult|renderBrowseItem' components/mushaf/MushafSearchView.tsx | sed -n '60,90p'
```

— L884 lands on the `<FlashList>` that consumes `renderBrowseItem` (annotated `FlashListRenderItemInfo<BrowseItem>`), and L998 lands on a `<FlatList>` consuming `renderSearchResult` (annotated `ListRenderItemInfo<SearchResultItem>` from `react-native`). The diff against `upstream/develop` also doesn't show the PR touching that `<FlatList>`, so it looks like it was a `<FlatList>` before the PR too.

Am I reading a different file state than you are? If `renderSearchResult` should be feeding a `<FlashList>` (and the L998 `<FlatList>` was meant to be migrated as part of a different change), happy to take that on as a follow-up — just want to make sure I understand before I swap the annotation. Apologies if I'm missing something obvious.

omar-zarka:
Rebased on latest `develop` (one trivial conflict in `components/collection/CollectionGrid.tsx` — your `palette-for-key` import landed alongside this PR's `LegendListRenderItemProps` import; resolved by keeping both).

On the annotation nit: I traced `renderSearchResult`'s actual consumer and `tsc` confirms it's wired to `<FlatList>` at line 997, not `<FlashList>`. I tried your suggested fix (annotate with `FlashListRenderItemInfo`) and `tsc` rejected it with the explicit \"Property 'target' is missing in type ListRenderItemInfo<...>\" error, since FlashList's `ListRenderItemInfo` requires `target` but RN's `FlatList` ListRenderItem callback signature doesn't include it. The existing `ListRenderItemInfo` from `react-native` is the correct annotation for this consumer.

I think line 1008 in your review may have referenced the `ListEmptyComponent` View one line below the `renderItem` prop (`<FlatList>` ends at line 1006+, then comes `ListEmptyComponent={<View ...>}` at line 1008). The actual `renderItem={renderSearchResult}` wiring is at line 999, inside `<FlatList>` opening at line 997. Sound right? Happy to look again if I'm misreading the structure.

All four other consumers (`CollectionView`, `CollectionGrid`, `SearchView`, `FavoriteRecitersSheet`) and the FlashList/RN alias split for `renderBrowseItem` are unchanged and correct per your earlier review.

CI: `tsc --noEmit` clean post-rebase.

osmansaeday (approved):
You're right, that was my mistake. The line number I cited (L1008) was indeed the `ListEmptyComponent` View, not the `renderItem` prop. `renderSearchResult` is wired to `<FlatList>` and the RN `ListRenderItemInfo` annotation is correct. Thanks for the careful trace plus the tsc-confirmed pushback.

Withdrawing the nit. PR is clean as-is. Approving + merging.

PR #277, opened 2026-05-23
chore: migrate remaining 13 fontMgr consumers to useMushafFontMgr hook (follow-up to #273)

omar-zarka:
Follow-up to #273. That PR introduced `useMushafFontMgr()` — a `useSyncExternalStore`-backed hook over the `mushafPreloadService.fontMgr` snapshot — and migrated the three Mushaf hot-path renderers (`SkiaPage`, `ContinuousMushafView`, `SurahDivider`) off the direct singleton read.

Osman's review on #273 flagged the remaining 12 components that still call `mushafPreloadService.fontMgr` at render time without subscribing. They risk a stale-null render if they mount before the preload completes — realistic windows for sheets opened via deep links into a verse action, settings screens opened from a tab other than Mushaf, etc.

## What this PR does

Migrates all 12 components Osman called out, plus a 13th caller spotted in passing (`SimilarVersesContent.tsx` — same shape, not on the original list), to `useMushafFontMgr()`:

- `components/MushafPageIcon.tsx`
- `components/ReadingThemeContent.tsx`
- `components/MushafSettingsContent.tsx`
- `components/sheets/VerseShareSheet.tsx`
- `components/sheets/SimilarVersesSheet.tsx`
- `components/sheets/verse-actions/ShareContent.tsx`
- `components/sheets/verse-actions/SimilarVersesContent.tsx` (bonus — same anti-pattern)
- `components/sheets/verse-actions/WBWContent.tsx`
- `components/share/SkiaVersePreview.tsx`
- `components/mushaf/qcf/QCFPage.tsx`
- `components/mushaf/reading/ReadingPageView.tsx`
- `components/mushaf/reading/ContinuousListView.tsx`
- `components/player/v2/PlayerContent/QuranView/index.tsx`

For the conditional-fontMgr sites (`isDK ? mushafPreloadService.fontMgr : null`), the hook is called unconditionally and its result is gated by the same condition — the subscription is what matters; the conditional just decides whether to consume it.

## Two non-migrations preserved

- `components/mushaf/main.tsx:853` reads `mushafPreloadService.fontMgr` inside a `navigateToPage` callback, not at render time. Callback-time reads pick up the current singleton snapshot; no subscription needed.
- Three reads of `mushafPreloadService.quranCommonTypeface` (in `MushafPageIcon`, `QCFPage`, `VerseShareSheet`, `ShareContent`) are kept as direct singleton reads. That field has no hook today; consumers that also use `useMushafFontMgr` will re-render when fontMgr changes, at which point the typeface read inside the recomputed `useMemo` will pick up the now-populated typeface.

## Depends on #273

This branch is stacked on `fix/mushaf-font-loading-race` (the branch behind #273) since it imports the new `useMushafFontMgr` hook from that PR. Once #273 merges, this PR's diff against `develop` will reduce to just the 13 consumer migrations.

## Test plan

- [x] prettier clean across all 13 files
- [ ] Cold-start the Mushaf tab on iOS sim — no regressions in pages, verse-action sheets, share sheet
- [ ] Open a verse-action sheet via a deep-link before the Mushaf tab has been visited — confirm the share preview + similar-verses surfaces render once preload completes (the case the migration is specifically targeting)
- [ ] Cold-start on Android emulator — same pass

### DIFF
--- a/components/MushafPageIcon.tsx
+++ b/components/MushafPageIcon.tsx
@@ -12,6 +12,7 @@ import {
 } from '@shopify/react-native-skia';
 import {moderateScale} from 'react-native-size-matters';
 import {mushafPreloadService} from '@/services/mushaf/MushafPreloadService';
+import {useMushafFontMgr} from '@/hooks/useMushafFontMgr';
 import {
   quranTextService,
   PAGE_WIDTH,
@@ -50,7 +51,7 @@ const MushafPageIcon: React.FC<MushafPageIconProps> = ({
   color,
   borderColor,
 }) => {
-  const fontMgr = mushafPreloadService.fontMgr;
+  const fontMgr = useMushafFontMgr();
   const uthmaniFont = useMushafSettingsStore(s => s.uthmaniFont);
   const mushafRenderer = useMushafSettingsStore(s => s.mushafRenderer);
   const fontFamily =
--- a/components/MushafSettingsContent.tsx
+++ b/components/MushafSettingsContent.tsx
@@ -22,6 +22,7 @@ import FormattedTextRenderer from '@/components/utils/FormattedText';
 import {LinearGradient} from 'expo-linear-gradient';
 import SkiaVerseText from '@/components/player/v2/PlayerContent/QuranView/SkiaVerseText';
 import {mushafPreloadService} from '@/services/mushaf/MushafPreloadService';
+import {useMushafFontMgr} from '@/hooks/useMushafFontMgr';
 import {digitalKhattDataService} from '@/services/mushaf/DigitalKhattDataService';
 import type {SkTypefaceFontProvider} from '@shopify/react-native-skia';
 import type {IndexedTajweedData} from '@/utils/tajweedLoader';
@@ -613,11 +614,12 @@ export const MushafSettingsContent: React.FC<MushafSettingsContentProps> = ({
     mushafRenderer === 'dk_indopak'
       ? 'DigitalKhattIndoPak'
       : mushafRenderer === 'dk_v1'
-      ? 'DigitalKhattV1'
-      : 'DigitalKhattV2';
+        ? 'DigitalKhattV1'
+        : 'DigitalKhattV2';
+  const subscribedFontMgr = useMushafFontMgr();
   const fontMgr =
     mushafPreloadService.initialized && digitalKhattDataService.initialized
-      ? mushafPreloadService.fontMgr
+      ? subscribedFontMgr
       : null;
 
   const actualTranslationText = useMemo(() => {
@@ -669,7 +671,10 @@ export const MushafSettingsContent: React.FC<MushafSettingsContentProps> = ({
         try {
           await digitalKhattDataService.switchRewayah('hafs');
         } catch (error) {
-          console.error('[MushafSettings] Failed to reset rewayah for QCF:', error);
+          console.error(
+            '[MushafSettings] Failed to reset rewayah for QCF:',
+            error,
+          );
         }
         setRewayah('hafs');
       }
@@ -861,9 +866,9 @@ export const MushafSettingsContent: React.FC<MushafSettingsContentProps> = ({
           <Text style={styles.settingRowLabel}>
             {themeMode === 'system'
               ? 'System'
-              : getReadingThemeById(
+              : (getReadingThemeById(
                   themeMode === 'light' ? lightThemeId : darkThemeId,
-                )?.name ?? 'System'}
+                )?.name ?? 'System')}
           </Text>
           <Feather
             name="chevron-right"
@@ -1139,15 +1144,17 @@ export const MushafSettingsContent: React.FC<MushafSettingsContentProps> = ({
             Rewayah switching is disabled in Mushaf 1440 beta.
           </Text>
         </View>
-      ) : hasDiffData(rewayah) && (
-        <RewayahDiffCard
-          rewayah={rewayah}
-          showRewayahDiffs={showRewayahDiffs}
-          toggleRewayahDiffs={toggleRewayahDiffs}
-          trackColor={trackColor}
-          styles={styles}
-          theme={theme}
-        />
+      ) : (
+        hasDiffData(rewayah) && (
+          <RewayahDiffCard
+            rewayah={rewayah}
+            showRewayahDiffs={showRewayahDiffs}
+            toggleRewayahDiffs={toggleRewayahDiffs}
+            trackColor={trackColor}
+            styles={styles}
+            theme={theme}
+          />
+        )
       )}
       <RewayahAccordion
         selectedId={rewayah}
--- a/components/ReadingThemeContent.tsx
+++ b/components/ReadingThemeContent.tsx
@@ -37,6 +37,7 @@ import {
   getQCFSurahNameChar,
 } from '@/constants/surahNameGlyphs';
 import {mushafPreloadService} from '@/services/mushaf/MushafPreloadService';
+import {useMushafFontMgr} from '@/hooks/useMushafFontMgr';
 import {
   digitalKhattDataService,
   BASMALLAH_TEXT,
@@ -190,7 +191,8 @@ export const ReadingThemeContent: React.FC<ReadingThemeContentProps> = ({
         : 'DigitalKhattV2';
   const fontsReady =
     mushafPreloadService.initialized && digitalKhattDataService.initialized;
-  const fontMgr = fontsReady ? mushafPreloadService.fontMgr : null;
+  const subscribedFontMgr = useMushafFontMgr();
+  const fontMgr = fontsReady ? subscribedFontMgr : null;
 
   return (
     <ScrollView
--- a/components/mushaf/qcf/QCFPage.tsx
+++ b/components/mushaf/qcf/QCFPage.tsx
@@ -54,6 +54,7 @@ import {
 } from '@/services/mushaf/QCFFontLoader';
 import {qcfLayoutCacheService} from '@/services/mushaf/QCFLayoutCacheService';
 import {mushafPreloadService} from '@/services/mushaf/MushafPreloadService';
+import {useMushafFontMgr} from '@/hooks/useMushafFontMgr';
 import {getTextAllahNameCharMap} from '@/services/mushaf/AllahNameHighlightService';
 import {themeDataService} from '@/services/mushaf/ThemeDataService';
 import {useMushafSettingsStore} from '@/store/mushafSettingsStore';
@@ -134,7 +135,7 @@ const QCFPage: React.FC<QCFPageProps> = ({
   onTap,
 }) => {
   const {theme} = useTheme();
-  const fontMgr = mushafPreloadService.fontMgr;
+  const fontMgr = useMushafFontMgr();
   const arabicTextWeight = useMushafSettingsStore(s => s.arabicTextWeight);
   const showThemes = useMushafSettingsStore(s => s.showThemes);
   const showAllahNameHighlight = useMushafSettingsStore(
@@ -361,7 +362,14 @@ const QCFPage: React.FC<QCFPageProps> = ({
       }
 
       qcfFontLoader.prefetch(
-        [pageNumber - 1, pageNumber + 1, pageNumber - 2, pageNumber + 2, pageNumber - 3, pageNumber + 3],
+        [
+          pageNumber - 1,
+          pageNumber + 1,
+          pageNumber - 2,
+          pageNumber + 2,
+          pageNumber - 3,
+          pageNumber + 3,
+        ],
         fontMgr,
       );
     })();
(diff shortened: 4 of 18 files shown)


osmansaeday (changes requested):
Thanks for tackling the 13-component migration. Reviewed the full diff. The migrations themselves are clean (more on that below), but there's an important dependency clarification to call out plus a blocker carried forward from #273.

## Dependency on #273

This PR is stacked on `fix/mushaf-font-loading-race`, not on `develop`. Against `develop` the diff includes all of #273's work inline: the new `hooks/useMushafFontMgr.ts`, the full refactored `services/mushaf/MushafPreloadService.ts`, AND the original 3-component migrations from #273. Net effect: merging #277 obviates #273, and #273 should be closed (or marked superseded) when this lands. Merging #273 first and then #277 would conflict on the service rewrite.

Treating #277 as "merge this, close that" for the rest of the review.

## Blocker

**HIGH (carried forward from #273, not fixed): `loadSkiaFontsSequentialFallback` unconditionally assigns `this._fontMgr = fontMgr`** at `services/mushaf/MushafPreloadService.ts:~236`.

Same failure mode as the #273 finding: when every `Skia.Typeface.MakeFreeTypeFaceFromData` returns `null` in the fallback loop (catastrophic device-level Skia failure), every iteration hits `continue`, the outer `try` succeeds, and `fontMgr` is a valid-but-empty `SkTypefaceFontProvider`. `_doInit`'s null-guard `if (!this._fontMgr)` evaluates to `false` (non-null object), `_state` becomes `'ready'`, `loadError` stays `false`, and `useMushafFontMgr`'s `if (loadError && !preloadFontMgr)` fallback branch never fires. Every component using the hook receives the empty provider, renders blank, no recovery.

The comment at the assignment site ("Publish whatever we got; even a partial provider keeps most of the Mushaf renderable") describes the partial-failure intent correctly but conflates partial failure (some families loaded) with total failure (zero families loaded). Same one-line fix from the #273 review: track `anyCriticalFailed` across the fallback loop, only assign `_fontMgr` when at least the critical DK family registered, leave `_fontMgr` null otherwise so the null-guard fires and the hook activates `useFonts`.

## Also still present from #273

**MEDIUM: two `useSyncExternalStore` calls per hook instance**. `hooks/useMushafFontMgr.ts:~634`. `preloadFontMgr` and `loadError` each subscribe with separate closures, so one `_notify()` schedules two re-renders per component per state transition. Under legacy rendering (Hermes default for most builds) this is a double-render every time the preload state changes. With 13+ components subscribing, this doubles render work at the busiest cold-start moment. Consolidate into a single store slice or derive `isReady` from one field since `loadError` and `fontMgr` always flip together at `_notify`.

**MEDIUM: `FALLBACK_FONT_SOURCES` comment misleadingly claims sync with `FONT_ASSETS`**. `hooks/useMushafFontMgr.ts:~612`. `FONT_ASSETS` has 8 entries, `FALLBACK_FONT_SOURCES` has 6 (Manrope intentionally excluded as UI fonts). The omission is correct, the comment is not. Future maintainer hazard. One-line fix: "Mushaf-critical subset of `FONT_ASSETS`; UI fonts (Manrope) intentionally excluded."

**LOW: retry doesn't reset partial state**. `initialize()` retry resets `_state` and `_initPromise` but leaves `_fontMgr`, `_quranCommonTypeface`, `_surahNameTypeface` at whatever the first run wrote. Harmless in practice (second run overwrites), but makes the invariant implicit. Explicit reset at the top of the retry block.

## Migration work is clean

Verified all 13 component migrations:

- `isDK` / `fontsReady` gates in `WBWContent`, `ContinuousListView`, `ReadingPageView`, `QuranView/index` are safe. `_state` flips to `'ready'` before `_notify()` in the `finally`, so when `useSyncExternalStore` re-reads on the triggered re-render, `.initialized` is already `true`. Not stale.
- `QCFPage` dep arrays correct: prefetch effect `[fontMgr, pageNumber]`, paragraph-build memo `[fontMgr, fontReady, pageLines, pageNumber, textColor]`. Null guard at L348 (`if (!fontMgr) return`) prevents pre-preload runs.
- `SkiaPage` and `ContinuousMushafView` `surahHeaderFonts` memo now lists `[lineWidth, fontMgr]` on both files. The MEDIUM dep-array bug from #273 is fixed in this PR.
- Direct `quranCommonTypeface` reads in `VerseShareSheet` and `ShareContent` are acceptable: the parent component re-renders via `useSyncExternalStore`, the typeface read picks up the populated value, passed directly as a prop. No memoization-stale-closure risk.
- `SkiaVersePreview`: clean swap, no `quranCommonTypeface` involvement.
- `useMushafFontMgr()` called inside `forwardRef`'s inner render function in `ContinuousListView` is Rules-of-Hooks compliant.

No new bugs from the migration itself. The block is entirely on the inherited service bug.

## Open questions

1. **Sequential-fallback design intent for total-failure case**: is "publish empty provider" intentional, or should zero-typefaces-loaded activate the `useFonts` fallback like a full failure does? The fix is mechanical (mirror the `anyFailed` tracking from the primary path), but the policy call is yours.
2. **Double-render tolerance under Hermes legacy rendering**: acceptable as known perf debt, or worth eliminating now given the 13+ hook subscribers?
3. **SSR / static-render paths**: both `useSyncExternalStore` calls return `mushafPreloadService.fontMgr` as the server snapshot, which is always `null` in Node. Any current or planned server-render path that hits these components (Remotion exports, OG-image gen)? If so, the fallback isn't enough on its own.

Happy to re-review on push. Once this lands cleanly, please close #273.

omar-zarka:
Thanks for the review @osmansaeday. New commit `77f5e5d` addresses all four findings.

## HIGH — empty-provider failure mode in sequential fallback

You're right that "publish whatever we got" silently conflates partial failure (some families) with total failure (zero families). Went with the more defensive variant: define an explicit `CRITICAL_FAMILIES` set (DK_V1/V2/IndoPak + QuranCommon) and track `anyCriticalFailed` across the loop. If any critical typeface fails OR zero typefaces register, `_fontMgr` stays null and the assignment is skipped — `_doInit`'s null-guard then flips `_state` to `'failed'` and `useMushafFontMgr`'s `useFonts` fallback fires per design.

SurahName fonts (V4/QCF) are treated as non-critical: their absence affects ornamental surah headers only, not the Mushaf body text, so dropping them still yields a usable provider. Codified the rationale in a comment at the assignment site so the next maintainer reading it understands the partial-vs-total split.

Open question (1) answered: zero-typefaces-loaded now activates `useFonts` like a full failure does. The mechanical fix mirrors `anyFailed` from the primary path; the policy call is "any critical-family loss → fall back."

## MEDIUM — double `useSyncExternalStore`

Consolidated to a single subscription returning a primitive boolean (`isReady`, defined as `_state === 'ready' || _state === 'failed'` via the existing `initialized` getter). `preloadFontMgr` and `loadError` derive from `mushafPreloadService.fontMgr` / `.loadError` at render time after the `isReady` flip. Both flip together inside `_notify`, so reading them post-`isReady` is consistent.

Primitive boolean snapshot means default identity comparison correctly short-circuits unchanged renders — no custom comparator needed. One `_notify()` now schedules one re-render per subscribing component per state transition.

Open question (2) answered: eliminated now. With 13+ subscribers the savings are meaningful at the busiest cold-start moment.

## MEDIUM — `FALLBACK_FONT_SOURCES` comment

Rewrote to: "Mushaf-critical subset of `FONT_ASSETS` in `MushafPreloadService`. UI fonts (Manrope) are intentionally excluded — they're loaded by the app-wide font loader, not the Mushaf preload service, so the fallback path doesn't need to re-load them. Updating this list does NOT need to mirror every `FONT_ASSETS` addition; only the Mushaf-rendering families."

## LOW — retry partial-state reset

`initialize()`'s `failed`-retry branch now explicitly nulls `_fontMgr`, `_quranCommonTypeface`, `_surahNameTypeface` alongside `_state` and `_initPromise`. The invariant "retry starts from a clean slate" is in the code now, not implicit on the second run's overwrites — defends against a sequential-fallback partial write being read mid-init.

## Open question (3) — SSR / static render

No current or planned server-render path for these Mushaf components. The app is RN-only; no Remotion exports, no OG-image gen against the Mushaf surface. If that changes, we'd want to lift the fallback decision out of the hook (so a server snapshot returning `null` doesn't render permanently blank) — but speculatively adding it would just add a code path we can't exercise. Happy to revisit if/when a static-render path appears.

## On #273

Confirmed: this PR carries all of #273's diff inline. Once #277 merges I'll close #273 as superseded. Merging in the opposite order would re-introduce the service-rewrite conflict.

## Validation

- `node node_modules/typescript/bin/tsc --noEmit` clean
- `node node_modules/prettier/bin/prettier.cjs --write` clean (no diff on either file)
- `eslint` clean on both files

Ready for re-review.

omar-zarka:
Thanks for the thorough re-review. All four findings addressed in `59a18b0` (force-pushed after rebase on latest `develop`):

**HIGH — empty-provider publish in `loadSkiaFontsSequentialFallback`** (`services/mushaf/MushafPreloadService.ts`)
Mirrored the `anyFailed` tracking from the primary path. The fallback loop now tracks `anyCriticalFailed` across the four critical families (DK_V1, DK_V2, IndoPak, QuranCommon). SurahName is treated as ornamental — its absence doesn't invalidate the provider. If any critical family fails to register, `_fontMgr` stays `null`, `_doInit`'s null-guard fires, `_state` flips to `'failed'`, and `useMushafFontMgr`'s `useFonts` fallback activates as designed.

**MEDIUM #1 — double-render from two `useSyncExternalStore` calls** (`hooks/useMushafFontMgr.ts`)
Consolidated to a single subscription returning a primitive boolean `isReady` (true when service `_state` is `'ready' || 'failed'`). `fontMgr` and `loadError` are now derived from `mushafPreloadService` at render time. One subscription = one re-render per state transition, regardless of how many components subscribe. Stable snapshot comparator since the getter returns a primitive.

**MEDIUM #2 — `FALLBACK_FONT_SOURCES` comment**
Rewrote the docstring: "Mushaf-critical subset of FONT_ASSETS; UI fonts (Manrope) intentionally excluded since they aren't loaded into the Skia typeface provider."

**LOW — retry partial-state reset**
`initialize()`'s failed-retry branch now explicitly nulls `_fontMgr`, `_quranCommonTypeface`, `_surahNameTypeface` alongside `_state` and `_initPromise`. Invariant is now explicit; no longer relies on second-run overwrites.

**Open questions answered:**

1. *Sequential-fallback total-failure intent* — agree it should activate the `useFonts` fallback rather than publish empty. The "publish whatever we got" comment described the partial-failure case correctly but conflated it with total failure. Fixed per the suggested guard.
2. *Double-render under Hermes legacy* — eliminating now. With the consolidated subscription it's no longer a concern regardless of bridgeless rollout timing.
3. *SSR/server-render paths* — none currently or planned in Bayaan or Qariah. The `useFonts` fallback is RN-only too, so even if a static-render path gets added later, both branches would need a shared null-guard at that point. Not blocking here.

Also closed #273 as superseded — `chore/migrate-fontmgr-consumers-to-hook` carries the full service rewrite + hook inline.

CI: `tsc --noEmit` clean, ESLint clean, Prettier clean. Happy to re-review on push.

osmansaeday (approved):
LGTM. All four findings cleanly addressed.

**HIGH (sequential fallback empty-provider)**: the `anyCriticalFailed || registered.size === 0` guard is correct. Walking the zero-typefaces-loaded case: every family fails → `registered.size === 0` → guard fires → `_fontMgr` stays null → `_doInit`'s null-guard sets `failed = true` → `_state = 'failed'` → `_notify()` → hook returns `fallbackFontMgr`. Recovery path is intact. The `CRITICAL_FAMILIES` constant cleanly defines what counts as a blocker (the 4 DK / QuranCommon typefaces, excluding Manrope which is UI-only).

**MEDIUM (two subscriptions)**: consolidated into a single `useSyncExternalStore` call with a primitive `boolean` snapshot. `Object.is` short-circuits correctly, so once `isReady` is `true` further `_notify()` calls produce no re-render. The 13+ subscriber × double-render storm at cold-start is eliminated.

**MEDIUM (`FALLBACK_FONT_SOURCES` comment)**: new docstring accurately describes the intentional Manrope exclusion and decouples future `FONT_ASSETS` additions from this list. Future-maintainer hazard removed.

**LOW (retry partial-state reset)**: explicit reset of `_fontMgr`, `_quranCommonTypeface`, `_surahNameTypeface` in the `'failed'` branch. Invariant is now explicit. Concurrent-caller race analysis holds (Caller 1 transitions `'failed' → 'idle' → 'loading'` synchronously; Caller 2 sees `'loading'` and falls through to the existing `_initPromise`).

**Migration verified intact across all 16 component files**. Both patterns (direct substitution vs guarded `initialized && dk_initialized` gate) are correct. The `surahHeaderFonts` memo `[lineWidth, fontMgr]` dep fix is in place on both `SkiaPage` and `ContinuousMushafView`. `ContinuousListView`'s hook call inside `forwardRef`'s render body is Rules-of-Hooks compliant. `useFonts` correctly consolidated into `useMushafFontMgr`'s `fallbackFontMgr`.

Two minor nits not worth another cycle:

1. **NIT**: `anyFailed` in the primary `loadSkiaFonts` is over-broad. It triggers the sequential retry if either Manrope font fails typeface creation, even though Manrope isn't in `CRITICAL_FAMILIES`. Spurious retry succeeds, so no correctness impact, just a minor cold-start delay on an already-rare failure mode. Consider mirroring `CRITICAL_FAMILIES` in the primary loop too.

2. **NIT**: `MushafPreloadService.ts` has two consecutive `/** ... */` JSDoc blocks before the class (one for the class, one for `MushafPreloadState`). Diff artifact, no runtime impact.

Open question answers reasonable across the board (zero-typefaces → fallback policy explicit, double-render eliminated, no SSR path in scope today).

Approving + merging. Once this lands, #273 stays closed (already done).

PR #279, opened 2026-05-24
docs(rfc-014): player scroll-handler seam — branding.useBottomSheetScrollableInQuranView

omar-zarka:
## Summary

Doc-only RFC proposing an optional `branding.useBottomSheetScrollableInQuranView?: boolean` field (default `true` — Bayaan's current behavior is byte-identical). When a fork sets it to `false`, `QuranView` skips `useBottomSheetScrollableCreator()` and passes no `renderScrollComponent` to `FlashList`. The swipe-down-to-close gesture on the player sheet is thereby opt-out; forks that need reliable imperative `scrollToIndex` calls (range-anchored seeks, bookmark deeplinks, future word-by-word follow-along) can trade the gesture for scroll correctness.

## Why doc-only first

Sprint 15 standing rule — unusual structural patterns deserve shape approval before code lands. Three options presented in the doc:

- **Option A** (recommended): boolean opt-out. Minimal surface, 1-line consumer change. Drawback: doesn't fully model the trade-off (what other scroll-strategy decisions might forks need?).
- **Option B**: factory slot — `branding.scrollComponentFactory?: () => RenderScrollComponent`. More flexible but allows any custom scroll wrapper, not just gorhom skip.
- **Option C**: auto-detect from RFC-013's `branding.initialPlayerVerseKey` presence. Most "magic" but couples two unrelated config fields.

Open question for review: which abstraction level is right, given the upstream's current single consumer + Qariah's known use case?

## Background

Sprint 23 (Qariah) added a two-piece workaround in `QuranView/index.tsx` — drop the gorhom `useBottomSheetScrollableCreator` opt-in AND add `key={currentSurah}` on the FlashList — to fix a partial-ayah seek bug that RFC-013's anchor enabled. Sprint 24's wholesale upstream-take of QuranView (RFC-013 absorption #269) dropped that workaround, surfacing the same partial-ayah regression on TestFlight 3.1.6 (1065). Re-applied as commits 56f77de + 5331148 on the Qariah fork; this doc-PR proposes lifting the workaround to a cross-fork config seam so future wholesale takes can't drop it again.

## Code PR scope (follows once shape clears)

- `config/branding.d.ts` — add the optional field.
- `components/player/v2/PlayerContent/QuranView/index.tsx` — read `branding.useBottomSheetScrollableInQuranView ?? true`; conditionally call the gorhom hook + pass `renderScrollComponent`. Bayaan default unchanged.
- Net diff ~10 lines.

## Test plan

- [ ] Bayaan default behavior unchanged (swipe-down-to-close still works on player sheet, scroll position resets to 0 on surah change).
- [ ] Qariah opts out via `branding.useBottomSheetScrollableInQuranView: false`; partial-ayah anchoring lands at the correct ayah on cold-mount + on surah change.
- [ ] No regression in regression tests (ssr / metadata).

### DIFF
--- a/docs/rfcs/014-player-scroll-strategy-seam.md
+++ b/docs/rfcs/014-player-scroll-strategy-seam.md
@@ -1,0 +1,434 @@
+# RFC-014 (XF-NNN draft): Player scroll-handler seam (`branding.useBottomSheetScrollableInQuranView`)
+
+**Status:** Draft (Qariah-side; doc-only PR following Sprint 15 standing rule — unusual structural
+patterns → shape approval before code lands)
+**Authors:** Omar Zarka (Qariah)
+**Targets:** `thebayaan/Bayaan` `develop`
+**Related:** RFC-013 (`thebayaan/Bayaan#269`, merged 2026-05-23 — `branding.initialPlayerVerseKey`)
+
+---
+
+## Summary
+
+Add an optional `branding.useBottomSheetScrollableInQuranView?: boolean` field (default `true`,
+preserving Bayaan's current behavior). When a fork sets it to `false`, `QuranView` skips
+`useBottomSheetScrollableCreator()` and passes no `renderScrollComponent` to `FlashList`. The
+swipe-down-to-close gesture on the player sheet is thereby opt-out; forks that need reliable
+imperative `scrollToIndex` calls (range-anchored seeks, bookmark deeplinks, future word-by-word
+follow-along) can trade the gesture for scroll correctness.
+
+---
+
+## Motivation
+
+### The cross-fork problem
+
+`QuranView/index.tsx` calls `useBottomSheetScrollableCreator()` from `@gorhom/bottom-sheet` and
+passes the resulting factory as `renderScrollComponent` on `FlashList`. This wires the Mushaf
+list's scroll gesture into the sheet's pan gesture recognizer so the user can swipe down anywhere
+on the Mushaf to dismiss the player — a natural and desirable UX for Bayaan.
+
+The gorhom integration has a well-known side effect: **it intercepts imperative scroll calls
+(FlashList `ref.current.scrollToIndex(…)` / `scrollToOffset(…)`) whenever the sheet's internal
+state machine has not yet settled into `EXTENDED` or `FILL_PARENT`**. During that transition
+window — which spans the full open animation on every player launch, and re-occurs on every sheet
+snap point change — any imperative scroll is silently no-oped. The list stays at its current
+position; no error is thrown, no callback fires.
+
+For Bayaan, which always starts playback from surah position 0 (ayah 1), the sheet transition
+window is irrelevant: the imperative `scrollToOffset({offset: 0})` in the surah-change effect
+lands at the same position the FlashList would already occupy, so the no-op is invisible.
+
+### Why Qariah hit it
+
+RFC-013 (`thebayaan/Bayaan#269`, merged 2026-05-23) added `branding.initialPlayerVerseKey(track)`
+so forks can anchor the Mushaf view on a specific ayah when a track starts playing. Qariah uses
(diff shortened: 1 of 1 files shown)


osmansaeday (approved):
Solid RFC. Motivation is real and well-evidenced (the 2-rAF deferral from RFC-013 lands inside the sheet's 300-450 ms open animation on iOS Release builds, so the gorhom wrapper's imperative-scroll interception breaks the partial-recitation anchor). The Sprint-15 doc-first convention is correctly followed, and the migration story for Qariah is clean.

Two substantive comments worth addressing before the code PR:

**1. Naming convention divergence.** Every prior `Branding` field is a noun phrase (`homeRowConfig`, `catalogVersionEndpoint`, `listenTabTopComponent`, `searchFilters`, `initialPlayerVerseKey`, `translationProvider`, `tafsirProvider`). `useBottomSheetScrollableInQuranView` is unique as a verb phrase, reads like a hook name. More importantly, a boolean closes the door on a future third strategy. If RN Gesture Handler v3 ever offers a different scroll-coupling pattern (or if Skia v2's gesture system grows one), you'd need a breaking-change second field. Recommend:

```ts
playerMushafScrollBehavior?: 'gorhom' | 'native';  // default 'gorhom'
```

String union extends cleanly: `'gorhom' | 'native' | 'gesture-handler-v3'`. Aligns with the noun-phrase convention. Resolves Open Question 3 in the RFC.

**2. Missed imperative scroll site (affects default Bayaan path, not just forks).** The RFC enumerates 3 imperative scroll call sites in `QuranView` (the 2-rAF deferred `scrollToIndex` for the RFC-013 anchor, the `scrollToOffset({offset: 0})` fallback, and the non-indexed-track `scrollToOffset`). It misses a 4th: the **active-ayah auto-scroll** that fires every time `currentVerseKey` changes while `isLocked === true`. This effect runs continuously during playback, including immediately after a player open when `isLocked` starts `true`. If the player is opened while the sheet is mid-animation, the first auto-scroll `scrollToIndex` is also intercepted by the gorhom wrapper, and the list stays at ayah 1 until the next `currentVerseKey` tick fires after the animation settles.

This is arguably a **worse UX than the RFC-013 partial-range case** because it hits the default Bayaan path. The RFC should:
- Acknowledge the 4th call site exists.
- Decide whether the `playerMushafScrollBehavior` flag also applies to it (probably yes), or whether it's deferred to a future RFC.

**Smaller suggestions:**

- "Default `true`" framing in the RFC header conflicts with how every prior RFC documents defaults. The pattern is: field absent from `branding.js` (Bayaan's default), consuming file has `?? fallback` inline. Reframe as "field absent in Bayaan's `branding.js`, `?? 'gorhom'` at the `QuranView` call site" to stay consistent.
- Android behavior is documented as "race is narrower, not observed." Worth adding one sentence confirming the flag applies identically on both platforms (a Qariah fork setting `'native'` would apply on Android too), so forks don't need platform-conditional branding.
- The "forks must provide an alternative dismiss affordance" requirement has no enforcement mechanism. A `__DEV__ && branding.playerMushafScrollBehavior === 'native' && console.warn(...)` in `QuranView` pointing to the RFC would surface the responsibility at dev time. Resolves Open Question 2 in the RFC.

**For Bayaan**: zero impact when the implementation PR lands. The field is absent from our `branding.js`, the `??` fallback resolves to gorhom, `renderScrollComponent` is passed unchanged. No `config/branding.js` change needed.

Approving the RFC. Reflect the two substantive comments in the doc before the implementation PR.

osmansaeday:
Closing in favor of #285, which ships the RFC doc (`docs/rfcs/014-player-scroll-strategy-seam.md`) inline with the implementation, matching the RFC-013 combined-doc-and-code precedent. The doc on #285 fully addresses the comments from this PR's review (naming adopted as `playerMushafScrollBehavior`, 4th scroll site enumerated and covered by the flag, `__DEV__` warning shipped). No content lost.

PR #280, opened 2026-05-24
chore(scripts): verify-version-sync — cover iOS extension targets

omar-zarka:
## Summary

Widens PR #251's `verify-version-sync.js` to also cover Xcode extension targets (Share Extension, Action Extension, Notification Service, Widget). Without this, extension target versions silently drift off the parent app's, producing Apple submission warnings (`CFBundleVersion of an app extension ('N') must match that of its containing parent app ('M')`).

## Why

iOS extension targets carry CFBundle values via Xcode build settings (`MARKETING_VERSION` + `CURRENT_PROJECT_VERSION` in `project.pbxproj`), not in their stub `*-Info.plist`. PR #251's verifier checked only the main app's `Info.plist` + Android gradle, so SE drift went undetected. On Bayaan today the SE is at 2.2.0/698 (stale since the version bumped) — this PR's `--fix` would patch it to the current 2.2.0/812.

## Implementation

- **`findExtensionConfigBlocks(repoRoot)`** — auto-discovers extension targets by scanning `project.pbxproj` for `XCBuildConfiguration` blocks whose `INFOPLIST_FILE` ends `-Info.plist` (Xcode's convention for non-main-app stubs). Anchored with `(?<![A-Z_])INFOPLIST_FILE` so `GENERATE_INFOPLIST_FILE = YES;` doesn't false-match.
- **`patchExtensionConfigBlocks(path, blocks, target)`** — rewrites `MARKETING_VERSION` + `CURRENT_PROJECT_VERSION` back-to-front so earlier blocks' byte offsets stay stable. Same idempotent semantics as `patchInfoPlist` / `patchAndroidGradle`.
- main() ties them in: extension mismatches participate in the iOS match boolean and the mismatch report, and `--fix` patches them too. Empty extension list (project with only a main app target) is a no-op — behavior unchanged for projects without extensions.
- 6 new unit tests on top of PR #251's existing 21 (4 for find, 2 for patch). Total 27/27 green.

## Test plan

- [x] Existing 21 tests still green (no regression on main-app path)
- [x] 6 new tests cover: empty-when-no-ios, empty-no-extensions, happy-path, regex-anchoring against GENERATE_INFOPLIST_FILE, patch-happy-path, main-app-untouched
- [x] End-to-end smoke against Bayaan's own pbxproj: SE detected at 2.2.0/698, `--fix` patches both Debug + Release blocks, re-verify reports ✓
- [x] `npx eslint` clean
- [x] `npx prettier --check` clean

### DIFF
--- a/scripts/__tests__/verify-version-sync.test.js
+++ b/scripts/__tests__/verify-version-sync.test.js
@@ -346,7 +346,7 @@ const SAMPLE_PBXPROJ = `// !$*UTF8*$!
 \t\t\t\tCURRENT_PROJECT_VERSION = 1;
 \t\t\t\tGCC_PREPROCESSOR_DEFINITIONS = (
 \t\t\t\t\t"DEBUG=1",
-\t\t\t\t\t"\$(inherited)",
+\t\t\t\t\t"$(inherited)",
 \t\t\t\t);
 \t\t\t\tGENERATE_INFOPLIST_FILE = YES;
 \t\t\t\tINFOPLIST_FILE = Bayaan/Info.plist;
@@ -430,7 +430,7 @@ describe('findExtensionConfigBlocks', () => {
   it('does not match GENERATE_INFOPLIST_FILE (regex anchoring)', () => {
     // Construct a config block that has GENERATE_INFOPLIST_FILE = YES but no
     // standalone INFOPLIST_FILE — should NOT be picked up as an extension.
-    const trick = `// !\$*UTF8*\$!
+    const trick = `// !$*UTF8*$!
 {
 \tobjects = {
 \t\tA1B2C3D4E5F6789012345678 /* Debug */ = {
--- a/scripts/__tests__/verify-version-sync.test.js
+++ b/scripts/__tests__/verify-version-sync.test.js
@@ -16,8 +16,10 @@ const {
   findInfoPlistPath,
   readVersionFromInfoPlist,
   readVersionFromAndroidGradle,
+  findExtensionConfigBlocks,
   patchInfoPlist,
   patchAndroidGradle,
+  patchExtensionConfigBlocks,
 } = require('../verify-version-sync');
 
 /**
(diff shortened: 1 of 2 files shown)


osmansaeday (approved):
LGTM on the logic. The `findExtensionConfigBlocks` + `patchExtensionConfigBlocks` pair is well-scoped: discovers `project.pbxproj` by scanning `ios/*.xcodeproj/`, extracts XCBuildConfiguration blocks whose `INFOPLIST_FILE` ends in `-Info.plist`, patches version keys back-to-front by byte offset. The integration into `main()` is clean (`extensionsMatch` checked before `allMatch`, fix path fully covered, post-fix re-verification includes extensions).

**Context correction**: my prior review on #251 said Bayaan "has no iOS extension targets today" - that was wrong. Bayaan already has a `ShareExtension` target with two XCBuildConfiguration blocks in `ios/Bayaan.xcodeproj/project.pbxproj` pointing at `ShareExtension/ShareExtension-Info.plist`. So this PR is immediately active for Bayaan, not just future-widgets defense. Current ShareExtension state: `MARKETING_VERSION = 2.2.0` / `CURRENT_PROJECT_VERSION = 698`, matching main app. The new guard will catch any drift on the next version bump.

**Findings worth addressing as follow-ups** (none blocking):

**IMPORTANT (test bug)**: the "no extension targets" test fixture is over-stripped. The regex that produces a main-app-only fixture removes the main-app UUID block alongside the extension blocks, leaving a pbxproj with no `XCBuildConfiguration` objects at all. `findExtensionConfigBlocks` returns `[]` because the outer regex never matches, not because it correctly skips the main-app `INFOPLIST_FILE = Bayaan/Info.plist` entry. If the `-Info.plist` suffix filter were dropped tomorrow, this test would still pass. Worth fixing: write the main-app-only block as a string constant rather than stripping from `SAMPLE_PBXPROJ`.

**SUGGESTION**: `blockRe` lazy terminator (`[\s\S]*?\n\s+\};`) matches the inner `buildSettings };`, not the outer `XCBuildConfiguration };`. Currently harmless because every key being read/written lives inside `buildSettings`, and the byte-offset patch still reconstructs correctly. But a future caller reading something outside `buildSettings` (e.g. `name = Release;`) will silently get a truncated block. Anchor the terminator more specifically.

**SUGGESTION**: extension blocks missing `MARKETING_VERSION` or `CURRENT_PROJECT_VERSION` are silently skipped via `if (!cpvMatch || !mvMatch) continue`. That code path is also hit by extensions that inherit version from the project level (Xcode 15+'s `GENERATE_INFOPLIST_FILE = YES` widgets, for instance). A `console.warn` on the skip would make the archive log show what was bypassed and why.

**SUGGESTION**: Widget targets using `WidgetName/Info.plist` (no dash, Xcode 15+ default) are excluded by the `-Info.plist` suffix filter. Documented limitation per the file comment, worth noting in the JSDoc on `findExtensionConfigBlocks` so the next contributor sees the constraint inline.

**Open question (separate cleanup PR coming)**: the main-app `XCBuildConfiguration` blocks in `project.pbxproj` show `MARKETING_VERSION = 1.0` (lines 514, 546) while `Info.plist` shows `2.2.0`. Looks like a stale pbxproj field never updated by `npm run version:*`. Not introduced by this PR. Opening a separate cleanup PR to bring those in sync.

**Bayaan-side**: the ShareExtension is already in sync. No immediate action. Will keep the `--fix` flag in mind for the next version bump.

Approving + merging.

PR #281, opened 2026-05-24
fix(recent-reciter-card): timeRemaining display for short / unknown-duration tracks

omar-zarka:
## Summary

Fix `RecentReciterCard#timeRemaining` returning `'0m'` in two distinct display states the user reads as "track ended":

1. **Duration unknown** — when `effectiveDuration` is 0 (duration hasn't been measured yet, or paused on a track with no `duration` field), the old code fell through to `'0m'`. Now returns `'—'` so the user knows the time is genuinely unknown rather than zero.

2. **Short tracks** — `Math.round(remainingSeconds / 60)` rounds to 0 for any remaining duration under 60s. Most visible on short surahs like Al-Fatihah (~50s total): every state past mid-track collapses to `'0m'`. Now shows seconds for remaining < 60s, e.g. `'20s'`.

Also tightens the "track ended" branch from `'0m'` → `'0s'` so the new lower-bound unit aligns with the new seconds branch above it.

## Discovery

Surfaced via user feedback on a downstream fork (Qariah's Greta walkthrough on Samsung Galaxy A55, TF-1036): user played Al-Fatihah for ~15s on the Listen tab → returning to Listen showed a "Continue Listening" card with `▶ 0m` badge even though ~35 seconds remained. Same `RecentReciterCard.tsx` lives on `develop` since its original Sprint-12-era introduction — no Qariah-specific patches involved.

## Test plan

- [x] `npx eslint` clean (no new warnings; the 2 existing warnings on lines 55/63 pre-date this change)
- [ ] Visual: short surah (e.g. Al-Fatihah) plays for ~15s, return to Listen, verify badge reads "20s" or similar (not "0m")
- [ ] Visual: long surah plays, behavior unchanged at minute resolution
- [ ] Visual: track in initial state with no duration → badge reads "—" not "0m"

Originally shipped on Qariah's fork as commit `e8df7d5` (Sprint 24 FB-4 fix). Cherry-picked here as a goodwill chore PR.

### DIFF
--- a/components/cards/RecentReciterCard.tsx
+++ b/components/cards/RecentReciterCard.tsx
@@ -111,20 +111,40 @@ export const RecentReciterCard = ({
   ]);
 
   // Calculate time remaining
+  //
+  // Two display states the old code conflated as `'0m'`:
+  //   1. **Duration unknown** — `effectiveDuration` is 0 because the
+  //      track's duration hasn't been measured yet, or we're paused on
+  //      a track with no `duration` field. The old `'0m'` reads as
+  //      "track ended" to the user. Now returns an em-dash so the user
+  //      knows the time is genuinely unknown rather than zero.
+  //   2. **Short tracks** — for any remaining duration under 60s,
+  //      `Math.round(remainingSeconds / 60)` rounds to 0 → `'0m'`. Most
+  //      visible on short surahs like Al-Fatihah (~50s total): every
+  //      state past mid-track collapses to "0m". Now shows seconds for
+  //      remaining < 60s, e.g. "20s".
   const timeRemaining = useMemo(() => {
     const effectiveDuration =
       isCurrentlyPlaying && storeDuration > 0 ? storeDuration : duration;
-    let remainingSeconds = effectiveDuration;
 
-    if (effectiveDuration > 0) {
-      if (isCurrentlyPlaying && storePosition >= 0) {
-        remainingSeconds = effectiveDuration - storePosition;
-      } else {
-        remainingSeconds = effectiveDuration * (1 - progress);
-      }
+    // Duration unknown → em-dash placeholder rather than '0m'
+    // which the user reads as "track ended".
+    if (effectiveDuration <= 0) return '—';
+
+    let remainingSeconds = effectiveDuration;
+    if (isCurrentlyPlaying && storePosition >= 0) {
+      remainingSeconds = effectiveDuration - storePosition;
+    } else {
+      remainingSeconds = effectiveDuration * (1 - progress);
     }
 
-    if (remainingSeconds <= 0) return '0m';
+    if (remainingSeconds <= 0) return '0s';
+
+    // Under a minute → show seconds so short surahs (Al-Fatihah / An-Nas
+    // / Al-Falaq / Al-Ikhlas) don't all collapse to "0m" near the end.
+    if (remainingSeconds < 60) {
+      return `${Math.max(1, Math.ceil(remainingSeconds))}s`;
+    }
 
     const totalMinutes = Math.round(remainingSeconds / 60);
     if (totalMinutes >= 60) {


osmansaeday (approved):
LGTM. Clean untangling of two separate bugs (unknown-duration tracks vs sub-60s short surahs) that both surfaced as `'0m'`. The early-return restructure (`effectiveDuration <= 0 → '—'` guard, then flat seconds vs minutes branches) is a clear readability win and correctly eliminates the dead `let remainingSeconds = effectiveDuration` assignment that was outside the old guard. Comment block is unusually thorough for a UI component, exactly the context that prevents future regressions.

Two minor nits, not blocking:

**NIT (consistency)**: rounding asymmetry between branches. `< 60s` uses `Math.ceil` (always rounds up), `>= 60s` uses `Math.round` (nearest). "20.4s remaining" shows "21s" but "20.4 minutes remaining" shows "20m". If the `ceil` in the seconds branch is intentional (never show "0s" while audio is still playing), worth a one-line comment so a future editor doesn't unify them.

**NIT (redundancy)**: `Math.max(1, Math.ceil(remainingSeconds))` — given the enclosing `if (remainingSeconds < 60)` only runs when `remainingSeconds > 0` (guaranteed by the earlier `effectiveDuration <= 0` early-return), `Math.ceil(positiveFloat)` is always >= 1. The `Math.max(1, ...)` guard is defensively correct but redundant.

**Edge case to verify** (not a blocker): if `storeDuration` or `duration` is ever `NaN`, `effectiveDuration <= 0` evaluates to `false` (NaN comparisons), so the code falls through to the `remainingSeconds` calc and renders `NaN s` or `NaN m`. `ExpoAudioProvider.tsx:75` computes `progress = position / duration` without a `duration <= 0` guard, so a transient NaN-duration is plausible. Pre-existing pattern, not introduced here, but a `Number.isFinite(effectiveDuration)` check alongside the `<= 0` guard would close the gap.

Bayaan-side: nothing.

Approving + merging.

osmansaeday (approved):
LGTM. Clean untangling of two separate bugs (unknown-duration tracks vs sub-60s short surahs) that both surfaced as `'0m'`. The early-return restructure (`effectiveDuration <= 0 -> '-'` guard, then flat seconds vs minutes branches) is a clear readability win and correctly eliminates the dead `let remainingSeconds = effectiveDuration` assignment that was outside the old guard. Comment block is unusually thorough for a UI component, exactly the context that prevents future regressions.

Two minor nits, not blocking:

**NIT (consistency)**: rounding asymmetry between branches. `< 60s` uses `Math.ceil` (always rounds up), `>= 60s` uses `Math.round` (nearest). "20.4s remaining" shows "21s" but "20.4 minutes remaining" shows "20m". If the `ceil` in the seconds branch is intentional (never show "0s" while audio is still playing), worth a one-line comment so a future editor doesn't unify them.

**NIT (redundancy)**: `Math.max(1, Math.ceil(remainingSeconds))`, given the enclosing `if (remainingSeconds < 60)` only runs when `remainingSeconds > 0` (guaranteed by the earlier `effectiveDuration <= 0` early-return), `Math.ceil(positiveFloat)` is always >= 1. The `Math.max(1, ...)` guard is defensively correct but redundant.

**Edge case to verify** (not a blocker): if `storeDuration` or `duration` is ever `NaN`, `effectiveDuration <= 0` evaluates to `false` (NaN comparisons), so the code falls through to the `remainingSeconds` calc and renders `NaN s` or `NaN m`. `ExpoAudioProvider.tsx:75` computes `progress = position / duration` without a `duration <= 0` guard, so a transient NaN-duration is plausible. Pre-existing pattern, not introduced here, but a `Number.isFinite(effectiveDuration)` check alongside the `<= 0` guard would close the gap.

Bayaan-side: nothing.

Approving + merging.

PR #270, opened 2026-05-23
chore: route brand strings through branding.appName

omar-zarka:
## What

Four user-facing literals hardcoded `"Bayaan"` — replace each with `${branding.appName}` so forks pick up their own name without having to fork the entire file. Default branding ships `"Bayaan"`, so output is byte-identical for the upstream app.

## Files touched

- `components/sheets/SurahOptionsSheet.tsx` — share copy (`"Listen to … on Bayaan"`)
- `components/reciter-profile/ReciterProfile.tsx` — share copy, both call sites
- `components/system-playlist/SystemPlaylistHeader.tsx` — `"Bayaan Curated"` badge
- `components/MushafSettingsContent.tsx` — Mushaf 1440 beta alert body

## Example diff

```diff
-    shareUrl(url, `Listen to ${surah.name} on Bayaan`);
+    shareUrl(url, `Listen to ${surah.name} on ${branding.appName}`);
```

## Why

Continuation of the RFC-007 cross-fork seams thread. These four files are the last small-surface holdouts where the fork has to maintain a whole-file copy just to swap a single literal — lifting through `branding.appName` retires that without changing default behavior.

## Test plan

- [x] `tsc --noEmit` clean.
- [x] No call-site behavior change for default branding (literal `"Bayaan"` returns from `branding.appName`).
- [ ] Confirm visually on next build (mechanical change; no runtime branching).

Mechanical change; no native build needed.

PR #283, opened 2026-05-24
chore(branding): convert top-level const arrays to getX() factories

omar-zarka:
## Summary

Three module-load-time const arrays still hardcoded "Bayaan" because they evaluate before any `branding` import resolves. The branding-conformance lint flags hardcoded brand strings in source but can't catch these — the literals live inside object initializers in a top-level array assignment, so a file's `import branding from '@/config/branding'` (if present) hasn't run yet when the array evaluates.

Surfaced in #270's defensive re-grep after the call-site brand-string sweep landed. Left unmigrated then because the fix is structural (`const X = [...]` → `function getX() { return [...] }`) rather than a one-line swap.

## Changes

| File | Symbol | Brand literal |
|---|---|---|
| `data/onboardingPages.ts:17` | `ONBOARDING_PAGES` → `getOnboardingPages()` | `'Welcome to Bayaan'` → `\`Welcome to ${branding.appName}\`` |
| `data/systemPlaylists.ts:287` | `SYSTEM_PLAYLISTS` → `getSystemPlaylists()` | `'Bayaan Essentials'` → `\`${branding.appName} Essentials\`` |
| `app/(tabs)/(d.settings)/credits.tsx:95` | `CREDITS` → `getCredits()` | `'Everyone who helped make Bayaan better'` → templated |

`systemPlaylists.ts`'s three internal helpers (`getSystemPlaylistById`, `getAllSystemPlaylists`, `getSystemPlaylistsByColumn`) now call the factory instead of referencing the deleted const; external consumers were already using these helpers so no further call-site changes needed.

`onboardingPages.ts`'s one caller (`components/modals/WhatsNewOnboarding/index.tsx`) updated at 4 sites: import + `useState` init + `setPages` call + `filterPagesByVersion` arg.

`credits.tsx`'s render wraps `getCredits()` in `useMemo(() => getCredits(), [])` since it's already a one-shot snapshot per render — same behavior as the prior top-level const, but now lazy.

## Out of scope

The `Bayaan team` footer literal at `credits.tsx:193` and the intro `Text` at `credits.tsx:149` are JSX text nodes (not const-array entries) — separate follow-up if desired. Same with any other prose mentions of "Bayaan" in JSX trees.

## Test plan

- [x] `npx eslint` clean (5 pre-existing warnings unrelated)
- [x] `npx prettier --check` clean
- [ ] WhatsNewOnboarding renders the same 14 pages with the same titles (the `branding.appName === 'Bayaan'` default means the rendered text is unchanged)
- [ ] System playlists grid renders the "Bayaan Essentials" tile (string also unchanged on the default branding)
- [ ] Credits screen renders all sections unchanged

Downstream confirmation: Qariah has been wanting this for two sprints — once this merges they can drop their downstream patches that special-case these three files.

### DIFF
--- a/app/(tabs)/(d.settings)/credits.tsx
+++ b/app/(tabs)/(d.settings)/credits.tsx
@@ -5,6 +5,7 @@ import {ScaledSheet, moderateScale} from 'react-native-size-matters';
 import {Feather} from '@expo/vector-icons';
 import Color from 'color';
 import {Theme} from '@/utils/themeUtils';
+import branding from '@/config/branding';
 
 interface Credit {
   title: string;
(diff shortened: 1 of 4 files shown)


osmansaeday (approved):
LGTM on the structural fix. Converting the three module-level const arrays to `getX()` factories correctly defers evaluation so that fork builds that override `config/branding.js` get the right values (the old const-at-import pattern baked "Bayaan" into the bundle before `branding.appName` resolved). The four touched files all hold together: every internal caller is updated, no dead imports remain. CI green.

**One MEDIUM worth a product call** (not blocking, but worth a follow-up):

`data/systemPlaylists.ts:298` — the `title` is correctly templated to `` `${branding.appName} Essentials` ``, but the `id: 'bayaan-essentials'` is still hardcoded. That `id` is:
- The URL path segment: `/system-playlist/bayaan-essentials`
- The PostHog analytics key for `playlist_play` events

So a Qariah-branded build will leak `bayaan-essentials` in deep-links and analytics events. Half-finished seam.

Two paths:
- **Stable internal key**: leave it as-is, accept the brand-leak in URLs/logs as a stable identifier (changing it later would break PostHog funnels and any existing deep-links).
- **Template the id too**: `` `${branding.appSlug}-essentials` ``. Cleaner branding but breaks any analytics history keyed on `'bayaan-essentials'` and any deep-link bookmarks pointing at the old URL.

Worth deciding before any fork ships with a different `appSlug`. For Bayaan today, no-op either way.

**LOW (fork-breaking, pre-existing risk)**: removing `ONBOARDING_PAGES` and `SYSTEM_PLAYLISTS` named exports without a backward-compat alias means any downstream fork that imports them directly (Qariah might) gets a TS compile error on pull. A one-release transition `export const ONBOARDING_PAGES = getOnboardingPages()` alias would soften the migration, but if Qariah's the only downstream consumer and they already know, this is fine.

**NIT**: `useMemo(() => getCredits(), [])` in `credits.tsx:129` — empty deps hide the factory's internal dependencies from `exhaustive-deps`. If anyone later adds a reactive value (e.g., `theme.colors.text`) inside `getCredits()`, the memo won't re-run. A one-line comment documenting the intentional `[]` would reduce future-editor risk.

**Bayaan-side action**: none required. `branding.appName === 'Bayaan'` means all rendered strings are unchanged for the upstream build.

Approving + merging.

osmansaeday (approved):
LGTM on the structural fix. Converting the three module-level const arrays to `getX()` factories correctly defers evaluation so that fork builds that override `config/branding.js` get the right values (the old const-at-import pattern baked "Bayaan" into the bundle before `branding.appName` resolved). The four touched files all hold together: every internal caller is updated, no dead imports remain. CI green.

**One MEDIUM worth a product call** (not blocking, but worth a follow-up):

`data/systemPlaylists.ts:298`, the `title` is correctly templated to `` `${branding.appName} Essentials` ``, but the `id: 'bayaan-essentials'` is still hardcoded. That `id` is:
- The URL path segment: `/system-playlist/bayaan-essentials`
- The PostHog analytics key for `playlist_play` events

So a Qariah-branded build will leak `bayaan-essentials` in deep-links and analytics events. Half-finished seam.

Two paths:
- **Stable internal key**: leave it as-is, accept the brand-leak in URLs/logs as a stable identifier (changing it later would break PostHog funnels and any existing deep-links).
- **Template the id too**: `` `${branding.appSlug}-essentials` ``. Cleaner branding but breaks any analytics history keyed on `'bayaan-essentials'` and any deep-link bookmarks pointing at the old URL.

Worth deciding before any fork ships with a different `appSlug`. For Bayaan today, no-op either way.

**LOW (fork-breaking, pre-existing risk)**: removing `ONBOARDING_PAGES` and `SYSTEM_PLAYLISTS` named exports without a backward-compat alias means any downstream fork that imports them directly (Qariah might) gets a TS compile error on pull. A one-release transition `export const ONBOARDING_PAGES = getOnboardingPages()` alias would soften the migration, but if Qariah's the only downstream consumer and they already know, this is fine.

**NIT**: `useMemo(() => getCredits(), [])` in `credits.tsx:129`, empty deps hide the factory's internal dependencies from `exhaustive-deps`. If anyone later adds a reactive value (e.g., `theme.colors.text`) inside `getCredits()`, the memo won't re-run. A one-line comment documenting the intentional `[]` would reduce future-editor risk.

**Bayaan-side action**: none required. `branding.appName === 'Bayaan'` means all rendered strings are unchanged for the upstream build.

Approving + merging.

PR #285, opened 2026-05-25
feat(branding): playerMushafScrollBehavior seam (RFC-014 code)

omar-zarka:
## Summary

Implements the RFC-014 player scroll-handler seam approved on #279. Adds `branding.playerMushafScrollBehavior?: 'gorhom' | 'native'` to the `Branding` type and consumes it in `QuranView`. Field absent from `config/branding.js` → `?? 'gorhom'` at the call site → byte-equivalent to today's behavior for Bayaan. Forks set `'native'` to opt out of the gorhom scrollable wrapper.

This PR supersedes the doc-only #279 — the RFC doc is included here in revised form to fold in Osman's review feedback.

## Changes folded in from the #279 review

- **Naming**: `playerMushafScrollBehavior: 'gorhom' | 'native'` (string union) instead of the originally-proposed boolean `useBottomSheetScrollableInQuranView`. Reasons:
  - Aligns with the noun-phrase convention of every prior `Branding` field (`homeRowConfig`, `catalogVersionEndpoint`, `listenTabTopComponent`, `searchFilters`, `initialPlayerVerseKey`, `translationProvider`, `tafsirProvider`).
  - Leaves room for a future third strategy (e.g. `'gesture-handler-v3'`, `'skia-gesture'`) without a breaking-change rename.
- **4th imperative scroll call site acknowledged.** The RFC now enumerates all four sites in `QuranView` (initialScrollIndex, the 2-rAF deferred surah-change scroll, the scroll-to-top fallback, and the active-ayah auto-scroll while `isLocked`). The seam covers all four uniformly.
- **`__DEV__`-only one-shot warning** at the call site when a fork picks `'native'`. Surfaces the "must provide an alternative dismiss affordance" responsibility. Module-scope guarded → fires once per JS context. Stripped from Release by `__DEV__` dead-code elimination.
- **Conditional spread** on FlashList's `renderScrollComponent` prop keeps shape parity with upstream's no-fork render path.
- **Platform parity** documented: single-valued across iOS and Android.

## What does NOT change in Bayaan

- `config/branding.js` — unchanged. The field is absent; the `?? 'gorhom'` fallback at the call site preserves the wrapper exactly as today.
- Runtime behavior of the player — byte-equivalent. The hook is still called unconditionally (rules-of-hooks), and the factory is still passed as `renderScrollComponent`.
- The swipe-down-to-dismiss gesture on the player sheet — unchanged.

## Reference implementation

Qariah's adoption ships in the same sprint, against the same shape. The Qariah-side commits provide a live reference for forks that want to opt into `'native'`.

## Test plan

- [ ] `tsc --noEmit` passes (verified locally).
- [ ] Manual: open the player on iOS Release simulator, verify swipe-down-to-dismiss still works (unchanged from `develop`).
- [ ] Manual: open the player on Android, verify swipe-down still works.
- [ ] Manual on a fork with `'native'`: verify imperative scrollToIndex lands consistently during the sheet's open animation.

## Cross-references

- Doc-only RFC PR: #279 (approved 2026-05-24 by @osmansaeday).
- RFC-013: #269 (merged 2026-05-23) — exposed the race this RFC fixes.
- Sprint 15 standing rule: "Unusual structural patterns → doc-only RFC PR first; code PR follows once shape clears."

### DIFF
--- a/components/player/v2/PlayerContent/QuranView/index.tsx
+++ b/components/player/v2/PlayerContent/QuranView/index.tsx
@@ -40,6 +40,24 @@ import branding from '@/config/branding';
 
 const surahData = require('@/data/surahData.json') as Surah[];
 
+// RFC-014 — module-scope one-shot dev warning for forks that opt into the
+// 'native' scroll behavior. Surfaces the "must provide an alternative
+// dismiss affordance" responsibility at dev time (no type-level
+// enforcement). Stripped from Release by the __DEV__ dead-code path at
+// the call site.
+let __warnedPlayerScrollBehaviorNative = false;
+function warnOncePlayerScrollBehaviorNative() {
+  if (__warnedPlayerScrollBehaviorNative) return;
+  __warnedPlayerScrollBehaviorNative = true;
+  console.warn(
+    "[branding.playerMushafScrollBehavior] === 'native': the player " +
+      "sheet's swipe-down-to-dismiss gesture is disabled at the Mushaf " +
+      'list level. Ensure an alternative dismiss affordance (e.g. an ' +
+      'explicit close button) is present in the player header. See ' +
+      'docs/rfcs/014-player-scroll-strategy-seam.md.',
+  );
+}
+
 // Module-scope header — stable component identity prevents FlashList unmount/remount
 interface QuranListHeaderProps {
   surahNumber: number;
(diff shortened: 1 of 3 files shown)


osmansaeday (approved):
LGTM. The implementation correctly resolves every comment from the #279 RFC review:

- **Naming adopted verbatim**: `playerMushafScrollBehavior: 'gorhom' | 'native'`, noun-phrase + string union extensibility for a future third strategy.
- **4th scroll site addressed**: the RFC doc now explicitly enumerates all four imperative scroll call sites and calls out the active-ayah auto-scroll (site 4) as "arguably the highest-impact one for default Bayaan". The flag covers all four uniformly rather than per-site opt-out.
- **`__DEV__` warning shipped**: module-scope one-shot guard with a specific, actionable warning text. Metro strips the whole block in Release via `__DEV__` dead-code elimination. Survives HMR cycles. Good pattern.

Spread shape on FlashList (`{...(renderScrollComponent ? {renderScrollComponent} : {})}`) is the cleaner of the two equivalent options. Rules-of-hooks satisfied (`useBottomSheetScrollableCreator()` called unconditionally, result conditionally consumed). No `any`, no unsafe casts.

Two LOW nits, neither blocking:

**NIT (docs)**: the `__DEV__` warning isn't mentioned in the `branding.d.ts` JSDoc for the `'native'` value. Fork authors who only read the type declaration won't know the warning exists. One-line addition to the JSDoc would complete the surface:

```typescript
 * `'native'` ... must provide an alternative dismiss affordance.
 *   A `__DEV__`-only console warning fires once per JS context when
 *   this value is active.
```

**NIT (cross-reference)**: `components/player/v2/TabletPlayer.tsx:412` also renders `QuranView` inside a gorhom bottom-sheet context. The seam works there today because `QuranView` reads `branding` directly (no prop drilling), but a future refactor that converts the branding read into a prop would silently miss the TabletPlayer call site. Worth a comment on the TabletPlayer line noting that RFC-014 applies, or a cross-reference in the RFC doc.

**Process note (non-blocker, already happened)**: the Sprint 15 standing rule is "doc-only RFC PR first; code PR follows." This PR ships both together. #279 (the doc-only RFC) is being closed as superseded by this PR. Worth keeping the doc-first rhythm for future RFCs to keep review surface tight.

**Bayaan-side**: zero changes. `branding.playerMushafScrollBehavior` stays unset, `?? 'gorhom'` falls through to current behavior, byte-equivalent to develop today.

Approving + merging.

PR #278, opened 2026-05-24
feat(timestamps): switch follow-along to R2 + mirror scripts + 2 new reciters

osmansaeday:
## Summary

Mobile side of the timestamps R2 mirror initiative. Pairs with [backend PR #18](https://github.com/thebayaan/bayaan-backend/pull/18).

**Plan doc:** `docs/superpowers/plans/2026-05-23-timestamps-r2-mirror.md`

### Runtime changes
- `TimestampFetchService` now fetches direct from `cdn.thebayaan.com/timestamps/{id}/{NNN}.json` instead of mp3quran/QDC at runtime. Single fetcher, no upstream API dependency, matches the audio resolution pattern.
- Follow-along registry gates on the new `has_timestamps` field from the backend response (no more probing every reciter).
- Dead mp3quran/QDC upstream types removed from `types/timestamps.ts`.
- `TimestampSource` union shrunk to `'r2' | 'local'`.

### New scripts
- `scripts/mirror-timestamps-to-r2.ts` and `scripts/lib/{r2-client, timestamp-sources}.ts`: one-shot R2 mirror, already ran (12,881 surah JSONs uploaded for 113 rewayat + Bandar Baleelah).
- `scripts/apply-timestamps-coverage.ts`: PATCHes `has_timestamps` + `timestamps_surah_list` onto each rewayat in the backend after the mirror.
- `scripts/strip-broken-qdc-ids.ts`: nulls `qdc_reciter_id` for 5 reciters whose QDC mappings returned empty verse_timings (Jibreel, Minshawi Mujawwad, Mustafa Ismail, Albanna, Maher Mujawwad 1440).
- `scripts/add-new-reciters.ts`: inserts 2 new mp3quran reciters via admin API (Ahmad Talib bin Humaid mp3quran read 137; Hasan Al-Daghriri mp3quran read 10905). Both will get their audio mirrored separately via the existing R2 audio worker, then their timestamps mirrored.

### Other
- `@aws-sdk/client-s3` moved from dependencies to devDependencies (it's only used by Node-side scripts, never imported by app code).
- `docs/features/ayah-timestamps.md` updated to reflect the R2 architecture.

## Dependencies

**Cannot fully take effect until backend PR #18 is merged + deployed to Railway.** The mobile changes are forward-compatible: until the backend exposes `has_timestamps`, the follow-along registry stays empty and the feature is just hidden (no crashes, no errors). After the backend ships, `apply:timestamps-coverage` populates the flags and follow-along activates.

## Test plan

[...]

PR #286, opened 2026-05-25
feat(branding): timestampCdnBase seam (RFC-015)

omar-zarka:
## Summary

Adds `branding.timestampCdnBase?: string` and consumes it in `TimestampFetchService` with an inline `??` fallback. Lifts the hardcoded `https://cdn.thebayaan.com/timestamps` URL introduced by #278 into a configurable seam — forks (Qariah included) can now mirror ayah-timestamp JSONs to their own R2 bucket without shadowing the entire fetch service.

Combined doc + code PR. Shape risk is low (single optional string field, same shape as RFC-010), and there's an active fork ready to consume the seam in the same sprint as the reference implementation.

## What does NOT change in Bayaan

- `config/branding.js` — unchanged. Field is absent; the `??` fallback at module load resolves `R2_BASE` to `https://cdn.thebayaan.com/timestamps`, byte-equivalent to today.
- Runtime behavior of the timestamps fetch — identical. Same URL composition, same headers, same error handling.

## Migration for forks

```js
// config/branding.js (fork's version)
module.exports = {
  // ... existing fields ...
  timestampCdnBase: 'https://cdn.myfork.example/timestamps',
};
```

That's it. The fork's mirror is responsible for serving the same `{base}/{rewayatId}/{NNN}.json` layout (already documented in #278's RFC text).

## Reference implementation

Qariah's adoption ships in the same sprint:

- `config/branding.js`: declares `timestampCdnBase: 'https://new.qariah-storage.xyz/timestamps'`.
- `services/timestamps/TimestampFetchService.ts`: consumes the seam (this PR's exact change).

## Test plan

- [ ] `tsc --noEmit` passes (verified locally).
- [ ] Manual: open the player on Zaynab's Mushaf (only reciter with timestamps on Bayaan today), verify the active-ayah highlight tracks audio — confirms `R2_BASE` still resolves to the same URL.
- [ ] Manual on a fork that overrides `timestampCdnBase`: verify the fetch hits the fork's CDN.

## Cross-references

- PR #278 — `feature/timestamps-r2-mirror` (merged 2026-05-22), introduced the hardcoded URL.
- RFC-010 — `branding.catalogVersionEndpoint` (analogous shape).

### DIFF
--- a/config/branding.d.ts
+++ b/config/branding.d.ts
@@ -219,6 +219,24 @@ export interface Branding {
    * See docs/rfcs/014-player-scroll-strategy-seam.md.
    */
   playerMushafScrollBehavior?: 'gorhom' | 'native';
+  /**
+   * RFC-015 — base URL for ayah-timestamp JSONs served from a fork's
+   * R2 bucket (or any CDN with the same
+   * `{base}/{rewayatId}/{NNN}.json` layout `TimestampFetchService`
+   * constructs).
+   *
+   * Field absent → consumer applies `?? 'https://cdn.thebayaan.com/timestamps'`
+   * at the call site → byte-equivalent to today's behavior.
+   *
+   * Forks set this to their own mirror's base when the `has_timestamps`
+   * flag is true for any reciter in their catalog and the timestamp
+   * JSONs live somewhere other than the Bayaan production CDN. No
+   * trailing slash; consumer composes
+   * `${timestampCdnBase}/${rewayatId}/${paddedSurah}.json`.
+   *
+   * See docs/rfcs/015-timestamp-cdn-seam.md.
+   */
+  timestampCdnBase?: string;
 }
 
 declare const branding: Branding;
--- a/docs/rfcs/015-timestamp-cdn-seam.md
+++ b/docs/rfcs/015-timestamp-cdn-seam.md
@@ -1,0 +1,210 @@
+# RFC-015: Timestamp CDN base seam (`branding.timestampCdnBase`)
+
+**Status:** Proposed (combined doc + code PR — small surface, single-line
+change at the consumer, low shape risk).
+**Authors:** Omar Zarka (Qariah)
+**Targets:** `thebayaan/Bayaan` `develop`
+**Related:** PR `thebayaan/Bayaan#278` (merged 2026-05-22) — timestamps R2
+mirror feature that introduced the hardcoded `cdn.thebayaan.com` URL base
+this RFC lifts.
+
+---
+
+## Summary
+
+PR #278 (timestamps R2 mirror) shipped with a hardcoded production CDN base
+URL in the timestamps fetch service:
+
+```ts
+// services/timestamps/TimestampFetchService.ts:5
+const R2_BASE = 'https://cdn.thebayaan.com/timestamps';
+```
+
+Forks that want to mirror ayah-timestamp JSONs to their own R2 bucket are
+forced to shadow the entire file today — there is no seam to swap just the
+URL base. This RFC adds an optional `branding.timestampCdnBase?: string`
+field; when absent, the consumer applies
+`?? 'https://cdn.thebayaan.com/timestamps'` and Bayaan's behavior is
+preserved byte-equivalent. When set, the consumer reads the fork's URL.
+
+The change is two files (`config/branding.d.ts` + `services/timestamps/TimestampFetchService.ts`)
+totaling under 15 lines. No new infrastructure, no API surface beyond a
+single optional string field. Same shape as RFC-010
+(`branding.catalogVersionEndpoint`) — combined doc + code PR is appropriate.
+
+---
+
+## Motivation
+
+A fork that wants to mirror timestamps to its own CDN today has three
+options, all bad:
+
+1. **Shadow `TimestampFetchService.ts`.** Every upstream change to the
+   file needs manual porting into each fork. The whole point of the
+   cross-fork seam work (RFC-007 onward) is to avoid this.
+2. **Patch the constant at build time** (e.g. a CI-side `sed`). Brittle,
+   invisible to fork maintainers reading the code, and a footgun for
+   anyone running the app locally.
+3. **Override the field on `branding` and pray.** Doesn't work — the
+   constant captures the literal at module load and never re-reads.
+
+A single optional config field absorbs all forks at one declaration site.
+
+### Why now
+
+Qariah-side timestamp mirroring is about to ramp up — TECH_DEBT #36
+(timing data for non-Zaynab reciters) tracks the work. Filing this RFC
+preemptively so the seam is in place before the first Qariah-side mirror
+upload happens, rather than racing the fork against a stuck-on-the-wrong-CDN
+release.
+
+The shape is small enough that a second fork can use the same seam without
+coordination overhead.
+
+---
+
+## Proposed change
+
+### `config/branding.d.ts` addition
+
+```typescript
+export interface Branding {
+  // ... existing fields ...
+
+  /**
+   * Base URL for ayah-timestamp JSONs served from a fork's R2 bucket
+   * (or any CDN with the same `{base}/{rewayatId}/{NNN}.json` layout
+   * `TimestampFetchService` constructs).
+   *
+   * Field absent → consumer applies
+   * `?? 'https://cdn.thebayaan.com/timestamps'` at the call site →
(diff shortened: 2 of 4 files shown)


osmansaeday (changes requested):
The seam is correctly shaped, Bayaan's runtime behavior is byte-equivalent (the `??` fallback to `https://cdn.thebayaan.com/timestamps` matches the previous hardcoded constant), and the combined-doc-and-code structure is right per the RFC-013 / RFC-014 precedent. Good RFC doc, good naming (`timestampCdnBase` follows the noun-phrase convention).

Requesting changes on one MEDIUM that's a fork-misconfig footgun, plus two LOWs / nits for awareness.

**MEDIUM (please fix before merge): no runtime guard against a trailing-slash `timestampCdnBase`**

`services/timestamps/TimestampFetchService.ts:11`. The JSDoc on the branding field says "no trailing slash" but there's no enforcement. A fork setting:

```js
timestampCdnBase: 'https://cdn.myfork.com/timestamps/'  // trailing slash, easy typo
```

produces composed URLs like `https://cdn.myfork.com/timestamps//rewayat-id/001.json`. Cloudflare R2 (and most CDNs) don't normalize double slashes, so this silently 404s every timestamp lookup, breaking follow-along for the fork with no useful error in the logs. Easy fix at the read site:

```ts
const R2_BASE = (branding.timestampCdnBase ?? 'https://cdn.thebayaan.com/timestamps').replace(/\/$/, '');
```

This is doc-says-X / code-doesn't-enforce-X drift; the documentation is the right contract, the runtime should match it.

**LOW (fork-only risk, pre-existing pattern but RFC widens the surface)**: `(await res.json()) as AyahTimestamp[]` at L56. Validates `Array.isArray` but no per-element shape check. If a fork's CDN serves a structurally wrong JSON (old format without `durationMs`, snake-cased fields from a raw mp3quran response uploaded by mistake), `writeTimestamps` writes `undefined` into the SQLite `duration_ms` column and the ayah-highlight logic gets `NaN` offsets. The cast existed pre-RFC, but this RFC explicitly widens the set of CDN operators to fork maintainers (untrusted relative to Bayaan), so the corruption surface grows. Consider a minimal zod-style shape check or at minimum an inline check that the first element has the expected keys before bulk-trusting the array.

**NIT (data lifecycle)**: `TimestampSource` enum changes from `'mp3quran' | 'qdc' | 'local'` on `develop` to `'r2' | 'local'` on the PR branch. User devices with cached `cached_surahs` rows will hold invalid enum values (`'mp3quran'`, `'qdc'`) until the next fetch overwrites them via `INSERT OR REPLACE`. Not a runtime crash (the source column is informational, not constrained at the SQLite layer), but worth either a one-shot migration on app startup or a comment in `TimestampDatabaseService` noting that stale values will linger until overwritten.

**Bayaan-side**: zero config change. The fallback URL is byte-equivalent to the pre-PR hardcoded constant.

**Open questions for you**:

1. Lint allowlist: the RFC mentions Qariah has a "branding-conformance lint allowlist" that needs entries for the fallback string. Is there a corresponding rule in Bayaan's lint config, and if so, does the allowlist entry need to be added in this PR?
2. SQLite migration: do you want a one-shot migration for the stale `source` enum values, or is overwrite-on-next-fetch fine?
3. Test surface: PR test plan says "open the player on Zaynab's Mushaf (only reciter with timestamps on Bayaan today)" - is that still the only entry with `has_timestamps: true`? If so the manual verification is narrow.

Requesting changes on the trailing-slash guard. Happy to re-approve on push; the rest can be follow-up.

omar-zarka:
Thanks for the careful review. Pushed `d229b87` addressing all three:

**MEDIUM — trailing-slash guard.** Done. `R2_BASE` is now `(branding.timestampCdnBase ?? '…').replace(/\/+$/, '')` at the read site. Used `/\/+$/` rather than `/\/$/` so accidental double-trailing-slash (`'…/timestamps//'`) is also normalized. Comment expanded to cite the failure mode (silent 404s on R2 / CDNs that don't normalize doubled slashes) so the next reader doesn't strip the guard as "defensive code that isn't doing anything".

**LOW — per-element shape check.** Done as a minimal inline guard rather than pulling in a runtime validator. `isAyahTimestampShape(raw[0])` probes the first element for the five required numeric fields before the `as AyahTimestamp[]` cast. If it fails, `console.warn` + return null — `writeTimestamps` never gets a chance to land `undefined` into `duration_ms`. Cheap, no dep, catches the dominant fork-CDN footgun (uploading a raw mp3quran response, an old shape without `durationMs`, snake-cased fields). Full per-element validation is overkill for the threat model (fork operators are mostly trusted; the surface that grew is "mis-upload your own CDN" rather than "adversarial CDN"); first-element probe is the 80/20.

**NIT — stale `'mp3quran'` / `'qdc'` enum values.** Added a block comment at the `CREATE TABLE cached_surahs` site. The column is informational (no CHECK constraint, never narrowed back to `TimestampSource` at any read site I could find), so overwrite-on-next-fetch is correct and a one-shot migration would be pure churn. Noted explicitly to spare the next reader the same archaeology.

**Answering your three open questions:**

1. **Lint allowlist.** Bayaan's `.eslintrc.cjs` doesn't carry a branding-conformance rule today (the lint scaffold for that lives in Qariah's fork). The fallback URL string `'https://cdn.thebayaan.com/timestamps'` only appears at the consumer site inside upstream Bayaan, so no Bayaan-side allowlist entry needed in this PR.
2. **SQLite migration.** Overwrite-on-next-fetch — `INSERT OR REPLACE` in `writeTimestamps` clears the stale row the first time the surah is re-fetched. Stale values never round-trip back to TS code as a `TimestampSource`, so the worst case is a row reads `'mp3quran'` in a debug query for one fetch cycle. The comment now documents this so it doesn't get re-investigated later.
3. **Test surface.** Confirmed narrow on Bayaan today — `grep '"has_timestamps":\s*true' data/reciters-fallback.json` returns zero rows. The R2 mirror feature (PR #278) shipped the wiring; Bayaan's catalog data hasn't been backfilled with `has_timestamps: true` for any reciter yet. Manual verification on the Bayaan side is effectively a "no-op path still works correctly" check — load any rewayat, confirm `hasSource` returns false and the timestamp fetcher is never invoked. Qariah-side verification will exercise the live path once we mirror Zaynab Talha (TECH_DEBT #36).

Re-requesting review.

osmansaeday (dismissed):
All three findings addressed cleanly.

**MEDIUM (trailing-slash)**: `.replace(/\/+$/, '')` is better than my suggested `/\/$/` (normalizes multiple-trailing-slash too, not just single). Comment expanded to cite the failure mode so the next reader doesn't strip it as cargo-cult defensive code. Verified at the read site of `R2_BASE`.

**LOW (per-element shape)**: `isAyahTimestampShape(raw[0])` first-element probe is the right call for the threat model (fork operators are mostly trusted, the surface that grew is "mis-upload your own CDN" rather than adversarial). Type guard structure is correct (returns `x is AyahTimestamp`, allows the subsequent `as` cast). Cheap, no dep.

**NIT (stale enum)**: comment-only at the `CREATE TABLE cached_surahs` site is the right tradeoff. Agreed that a one-shot migration is pure churn given the column is informational and never narrowed back to `TimestampSource`.

Open question answers all reasonable, especially the Bayaan-side narrow test surface confirmation (`grep '"has_timestamps":\s*true' data/reciters-fallback.json` returns zero rows). The Bayaan path is the no-op default for now.

Status note: branch is CONFLICTING/DIRTY against current `develop` (#285 added `playerMushafScrollBehavior` to `config/branding.d.ts` in the same `Branding` interface region where this PR adds `timestampCdnBase`). I'll rebase via maintainer-edits to unblock, since the conflict is a mechanical interleave of two optional keys.

Re-approving.

osmansaeday:
Rebased onto current `develop` (post-#285 merge) for you via maintainer edits. Conflict was a clean interleave in `config/branding.d.ts`, both `playerMushafScrollBehavior` (RFC-014) and `timestampCdnBase` (this PR) add optional keys to the end of the `Branding` interface. Resolution: kept both keys, RFC-014 first then RFC-015. No code logic changed. CI is now re-running on `2d06159`. Will re-approve and merge once green.

osmansaeday (approved):
Re-approving after the maintainer-edit rebase. branding.d.ts merge resolution is mechanical (interleaved RFC-014 + RFC-015 optional keys, both already-reviewed content). Will merge when CI green.

PR #289, opened 2026-05-25
feat(rfc-016): server-driven home row config (doc + client)

osmansaeday:
## Summary

Implements RFC-016: moves the Listen-tab home row order from the bundled `branding.homeRowConfig` (PR #260) to a backend endpoint `GET /v1/home-config` (thebayaan/bayaan-backend#23). Same wire format consumed identically by every client platform: mobile (this repo), web (existing), TV (in progress on `feature/tv-platforms`).

Builds directly on @omar-zarka's PR #260. The 12-row `HomeRowId` union and `RecitersView`'s `filter+map` rendering at `components/RecitersView.tsx:556-748` are unchanged. The only structural client change is at `components/RecitersView.tsx:726`:

```diff
- const homeRowConfig = branding.homeRowConfig ?? DEFAULT_HOME_ROW_CONFIG;
+ const remoteRows = useRemoteHomeConfig();
+ const homeRowConfig = remoteRows ?? branding.homeRowConfig ?? DEFAULT_HOME_ROW_CONFIG;
```

Bundled `branding.homeRowConfig` stays as the offline fallback. Server cannot ship unknown ids that crash the client (the existing `rowNodes[row.id] ?? null` skip handles forward-compat).

## What's in this PR

1. **`docs/rfcs/016-server-driven-home-rows.md`** — the RFC document.
2. **`hooks/useRemoteHomeConfig.ts`** — MMKV-cached hook that fetches `/v1/home-config` on cold-mount + AppState `active` (5-min debounce, 1.5-s timeout, fail-open).
3. **`components/RecitersView.tsx`** — one-line swap to prefer remote rows.

## Backend half

Lives in a separate PR: **thebayaan/bayaan-backend#23**. Adds the `home_config` table + `GET /v1/home-config` route + idempotent migration seeding version 1 with today's 12-row order, so first deploy is verbatim parity.

Both PRs should land together, but mobile is safe to merge first because the hook falls back to the bundled config when the endpoint is missing or 404s.

## Test plan

- [x] `npx tsc --noEmit` — clean.
- [x] `npx prettier --write` on changed files.
- [x] Backend PR: `bun test` 220/220 pass (4 new tests cover the endpoint).
- [ ] Smoke-test in dev: hook returns remote rows when the dev backend is up; falls back to bundled when env vars are missing.

## Maintainer ask (cc @omar-zarka)

Two specific things I want your read on:

[...]

PR #291, opened 2026-05-26
docs(rfc-017): server-driven config envelope (responds to RFC-016 maintainer ask)

omar-zarka:
## Summary

Codifies the wire-shape and client-side polling contract that RFC-010 (`/v1/catalog-version`) and RFC-016 (`/v1/home-config`) both consume into a single named pattern: the **server-driven config envelope**. Doc-only. Zero code change.

Directly responds to the two specific questions in [RFC-016's maintainer-ask section](https://github.com/thebayaan/Bayaan/pull/289):

1. **No \`?platform=\` from day 1** — endorsed. The escape hatch (client-owned renderer registry + new id for platform-only features) is named explicitly so the next contributor does not re-litigate.
2. **Wire shape \`{version, updated_at, ...payload}\`** — formalized as the envelope. Four candidate future consumers listed (Listen-tab top slot variant, search-filter dimensions, onboarding row order, featured-playlists surface) to show the contract is paying for itself across 4–6 future seams.

Also codifies the client-side polling contract (1500 ms timeout, AppState-debounced foreground refresh, fail-open, \`server.version > cache.version\` not \`!==\`, bundled fallback required) that both RFC-010 and RFC-016 already implement — so future seams inherit it in two lines instead of re-deriving.

## What this PR contains

One new file: \`docs/rfcs/017-server-driven-config-envelope.md\` (+191 lines). No code, no schema change.

RFC-010 and RFC-016 endpoints are conformant by construction; neither needs amending.

## Maintainer ask

cc @osmansaeday. If you accept this RFC, [#289](https://github.com/thebayaan/Bayaan/pull/289) (RFC-016) can ship as-is and future RFCs of the same family cite this one. If you want adjustments to the envelope before locking it in, this is the cheapest moment to make them.

Open questions parked in the RFC's Open Questions section (\`min_client_version\` reservation, \`updated_at\` required vs optional, bundled-fallback location convention) — happy to resolve any before merge.

Happy to fold this into #289 directly if you would prefer one document instead of two; the argument for keeping it separate is that the envelope contract outlives any single seam.

## Test plan

- [ ] @osmansaeday reviews RFC content + open questions
- [ ] Confirm RFC-017 is the next available number (014/015/016 currently in flight on separate branches; check no collision against any local in-flight RFC numbering)


osmansaeday:
### Code review

Found 3 issues:

1. The envelope spec contradicts the shipped RFC-016 implementation it cites as conformant. The RFC says the payload key is "always at the top level (not nested under `data` or `payload`)" and that the existing endpoints are "conformant by construction", but `useRemoteHomeConfig.ts` parses `{ data: { version, updated_at, rows } }` and reads `body.data`, which is exactly the shape rejected as Alternative B. One of the two needs to change before this contract is locked in.

https://github.com/thebayaan/Bayaan/blob/252c197ea60c1459fa74d9032e70f1a6e9e7cbce/docs/rfcs/017-server-driven-config-envelope.md#L69-L71

https://github.com/thebayaan/Bayaan/blob/85de88238a2e2e1665dc5b1b6d49431a2b27a229/hooks/useRemoteHomeConfig.ts#L30-L36

https://github.com/thebayaan/Bayaan/blob/85de88238a2e2e1665dc5b1b6d49431a2b27a229/hooks/useRemoteHomeConfig.ts#L80-L83

2. The endpoint-conventions table misstates the shipped backend on three rows: `/v1/home-config` requires a Bearer key via the `apiAuth` middleware (table says "Public (no auth) for read"), `homeConfigService` throws `NotFoundError` (HTTP 404) when no row exists (table mandates 200 with an empty payload), and the route sets no Cache-Control header (table mandates `public, max-age=60, s-maxage=60`). Given the "conformant by construction" claim, either the backend or the table needs reconciling. Backend refs: `src/routes/v1/home-config.ts`, `src/middleware/apiAuth.ts`, `src/services/homeConfigService.ts`.

https://github.com/thebayaan/Bayaan/blob/252c197ea60c1459fa74d9032e70f1a6e9e7cbce/docs/rfcs/017-server-driven-config-envelope.md#L76-L82

3. The client-side polling contract omits the payload shape-validation step that both reference implementations perform and that the RFC-015 review (PR #286) established for fork-facing payloads. A future seam following the six steps verbatim would bulk-trust `res.json()`. Worth adding as an explicit step since this doc is meant to be cited instead of re-derived.

https://github.com/thebayaan/Bayaan/blob/252c197ea60c1459fa74d9032e70f1a6e9e7cbce/docs/rfcs/017-server-driven-config-envelope.md#L85-L96

https://github.com/thebayaan/Bayaan/blob/85de88238a2e2e1665dc5b1b6d49431a2b27a229/hooks/useRemoteHomeConfig.ts#L83-L89

omar-zarka:
Thanks Osman — all three were the doc lying about the shipped code, not the code being wrong. Reconciled the RFC to describe what `/v1/home-config` + `useRemoteHomeConfig.ts` actually do (pushed as `46a1454`, doc-only). Point by point:

**1. Envelope shape contradicted the shipped `data` wrapper.** You're right — `useRemoteHomeConfig.ts` declares `interface RemoteConfigResponse { data: { version, updated_at, rows } }` and reads `const next = body.data`, which is exactly the shape I'd labelled "Alternative B" and rejected. The doc now adopts the **`data`-wrapped envelope** (`{ data: { version, updated_at, ...payload } }`) as the canonical shape:
- "The envelope" section + both JSON examples are now `data`-wrapped, with the home-config example annotated as the exact shape the hook parses.
- The conventions/table fields are `data.version` / `data.updated_at` / `data.{payload}`.
- The old "always top-level, not nested under `data`" claim is gone. Alternative B is now the *flat* envelope (rejected because re-flattening would require a backend change + coordinated client migration for a cosmetic win). So "conformant by construction" is now literally true for the deployed code.

**2. Conventions table misstated the backend on three rows.** Corrected all three to match what you describe as shipped (refs: `routes/v1/home-config.ts`, `middleware/apiAuth.ts`, `services/homeConfigService.ts`):
- **Auth** → "Bearer API key required via the existing `apiAuth` middleware" (was "Public (no auth)"). Matches the hook sending `Authorization: Bearer ${EXPO_PUBLIC_BAYAAN_API_KEY}`.
- **Missing-row status** → `404` (`NotFoundError` thrown by `homeConfigService`), with the prose reworked: the client's fail-open path already treats `404` correctly (non-`2xx` → keep cached/bundled). The "deliberately cleared" signal is now an explicit `"rows": []` row returning `200`, not a `404`. (was: "mandate 200-empty, never 404").
- **Cache-Control** → "Not set today" + the `max-age=60, s-maxage=60` line demoted to an explicit **recommendation / follow-up, not a conformance claim**. Also fixed the downstream `s-maxage` risk bullet and the Alternative-C aside that had asserted the CDN cache as already-shipped.

**3. Polling contract omitted the shape-validation step (per the PR #286 / RFC-015 review).** Added it as an explicit step in the six-step sketch (now seven steps): after `res.json()`, unwrap `body.data` and validate before trusting it — reject as fail-open if `data` is absent, `data.version` isn't a `number`, or the payload isn't the expected type (`Array.isArray(data.rows)`). Quoted the hook's actual guard verbatim and cross-referenced the RFC-015 fork-facing-payload requirement, so a future seam copying the steps can't bulk-trust `res.json()`.

Net: the RFC now describes the deployed reality, and the only forward-looking item (the CDN `Cache-Control` header) is flagged as a recommendation rather than a claim. Happy to adjust any of the three if you'd rather the contract move in a different direction.

osmansaeday (commented):
The design is right (a separate RFC is the correct call, and `version > cache` over `!==` is well-justified for the rollback case), and I verified the polling-contract values against `useRemoteHomeConfig.ts` (1500ms timeout, `>` comparison, shape validation, fail-open, all match). But the load-bearing **"conformant by construction"** claim is half-true and should be reconciled before this is stamped as canonical:

- **RFC-010's reference impl is flat, not `data`-wrapped.** `catalogVersionPoll.ts:76-77` reads `body.version` directly; there's no `data` envelope. So "both RFC-010 and RFC-016 already unwrap `body.data`" is inaccurate for RFC-010.
- **RFC-016 disagrees with itself.** The shipped hook reads `body.data`, but RFC-016's own merged doc shows the flat `{version, updated_at, rows}` shape (and "no auth, public") while the hook sends a Bearer token. The code already diverged from its RFC.

Net: scope the conformance claim to the RFC-016 *hook*, explicitly grandfather/migrate RFC-010's flat shape, and fix or annotate the RFC-016 doc-vs-code skew (`body.data` + Bearer).

Also tighten two contract details: step 2 says debounce on "last *successful* fetch" but both impls debounce on last *attempt*; and the required bundled fallback is actually a two-level chain (`branding.homeRowConfig ?? DEFAULT_HOME_ROW_CONFIG`), so name which layer is the guaranteed one.

Nit: the "likely RFC-018/019" predictions are already taken by unrelated PRs, so drop the numbers.

omar-zarka:
Reconciled in `ab9f2aa` (doc-only):

- Scoped "conformant by construction" to the RFC-016 **hook**, and added a callout that **grandfathers** RFC-010's flat `body.version` shape (it reads flat, not `data`-wrapped) — new consumers must use the `data` wrapper, RFC-010 is the only exemption. Corrected the sentence that previously claimed both impls "already unwrap `body.data`."
- Flagged the RFC-016 doc-vs-code skew (the doc shows flat `{version, updated_at, rows}` + "no auth, public", while the shipped hook reads `body.data` and sends a Bearer) and named the deployed code as source of truth, with a request to reconcile the RFC-016 doc.
- Debounce wording → "last fetch **attempt**" (both impls reset the timer on a failed fetch).
- Named the two-level bundled fallback (`branding.homeRowConfig ?? DEFAULT_HOME_ROW_CONFIG`) and called out the upstream `DEFAULT_HOME_ROW_CONFIG` as the guaranteed layer.
- Dropped the RFC-018/019 number predictions.

osmansaeday (approved):
Re-review: the conformance claim is now reconciled. The added "Conformance scope" block scopes "conformant by construction" to the RFC-016 hook, explicitly grandfathers RFC-010's flat `body.version` shape (and flags it should migrate to the wrapper when it next changes), and calls out the RFC-016 doc-vs-code skew (`body.data` + Bearer) for reconciliation. The polling contract now says debounce on last *attempt* (matching both impls), the bundled fallback names `DEFAULT_HOME_ROW_CONFIG` as the guaranteed layer with `branding.homeRowConfig` as the optional override, and the stale RFC-018/019 number guesses are dropped. LGTM.

PR #292, opened 2026-05-29
chore(translations): guard against simultaneous downloads in Translations & Tafseer

omar-zarka:
## Problem

`handleDownload` / `handleDownloadTafseer` in `TranslationsContent.tsx` invoke the store action without surfacing any feedback when a download of that type is already in flight. The stores already guard against a true double-start internally (`translationStore` and `tafseerStore` both `if (downloadingId) return` before doing any work), so there is **no corruption risk** — but the second tap is silently ignored at the store level, which reads as an unresponsive button.

## Fix

Add a per-handler guard on that download type's **own** in-flight id (`downloadingId` for translations, `tafseerDownloadingId` for tafseer). When busy, show an Alert ("Download in progress, …") and return; otherwise proceed to the existing `try { await download… }`.

Translations and tafaseer write to **separate** SQLite DBs (`translations.db` / `tafaseer.db`) with no shared write path, so a translation download and a tafseer download can safely run concurrently — each handler guards only its own state (no cross-blocking).

`useCallback` deps updated to match (`[…, downloadingId]` / `[…, tafseerDownloadingId]`).

## Why

The value here is **user-facing feedback**, not a correctness fix — a rapid second tap on the same download type now gets a clear Alert instead of being silently no-op'd at the store level.

## How discovered

Surfaced on the Qariah fork during user testing (a tester tapped a download row twice and got no visible response). The same component exists upstream verbatim; mirroring with no fork-specific behavior.

## Test plan

- [ ] Settings → Translations & Tafseer
- [ ] Tap a translation download; while in progress, tap another translation → Alert fires
- [ ] Tap a tafseer download; while in progress, tap another tafseer → Alert fires
- [ ] **Concurrent across types is allowed:** start a translation download, then a tafseer download → both proceed (no cross-block)
- [ ] Failure case unchanged: with no download in progress, force a network failure → original "Download Failed" Alert still fires

## Notes

- Pure UX guard — no store changes, no new dependencies.
- The Alert title ("Download in progress") is sentence-case matching the rest of the file's Alert titles.
- `lightHaptics()` stays at the top of each handler — the tap is acknowledged even when guarded.

---

**Correction (per review):** the real value here is the **Alert feedback**, not corruption prevention. Both stores already no-op a rapid double-tap at the store level (`translationStore` / `tafseerStore` early-return on an in-flight id), so there is no data-corruption risk to fix — this PR just makes that silent refusal visible. The guards are per-download-type (each handler checks only its own store's in-flight id); translations and tafaseer write separate DBs and run concurrently.


OKKHALIL3 (dismissed):
## Review

The UX improvement here is real and worth landing — surfacing an Alert when a download is already in progress is better than the current silent no-op. But there are two issues that need addressing before this merges.

---

### Cross-guard is unjustified

`handleDownload` and `handleDownloadTafseer` both check `downloadingId || tafseerDownloadingId`, which means a running translation download blocks tafseer downloads and vice versa.

The PR description justifies this with "they share the same network/cache pressure" — but looking at the DB layer, translations write to `translations.db` and tafaseer write to `tafaseer.db` (separate SQLite files, separate connections, separate tables). There is no shared write path between them. Running both concurrently is safe.

The cross-blocking is pure UX friction with no correctness benefit. A user who starts a translation download and then goes to download a tafseer gets an unexpected rejection for no reason.

Fix: each handler should guard on its own state only.

```ts
// handleDownload
if (downloadingId) { ... }

// handleDownloadTafseer
if (tafseerDownloadingId) { ... }
```

---

### The bug description overstates the severity

The PR frames this as a data corruption risk ("concurrent fetches can corrupt or partially-overwrite each other"). Worth noting that both stores already guard against this silently — `translationStore` line 27 and `tafseerStore` line 35 both have `if (downloadingId) return` before doing any work. A second rapid tap already no-ops at the store level with no corruption.

The real value of this PR is the Alert feedback, not the correctness fix. The description should reflect that so reviewers aren't misled about the severity.

---

Everything else looks fine — the `useCallback` dependency updates are correct, the Alert copy matches the file's existing style, and `lightHaptics()` firing before the guard is consistent with every other handler in this component.

omar-zarka:
Thanks @OKKHALIL3 — both addressed in a8c1b4c:

1. **Dropped the cross-guard.** You're right that `translations.db` and `tafaseer.db` are separate with no shared write path, so concurrent is safe. Each handler now guards on its own id only (`downloadingId` / `tafseerDownloadingId`); deps updated to match.
2. **Reframed the description.** Removed the corruption/severity framing and noted that both stores already silently no-op a double-tap — the real value here is the user-facing Alert feedback, not a correctness fix.

The Qariah-side copy got the same own-state fix. Re-requesting review.

omar-zarka:
@OKKHALIL3 this is ready for another look — the cross-guard was dropped in a8c1b4c (each handler now guards only its own id, since translations.db and tafaseer.db have no shared write path). Re-requesting review when you have a moment.

osmansaeday (commented):
Architecturally sound: guarding each handler on its own store's in-flight id keeps the source of truth in the store, the `useCallback` deps are correctly updated (no stale closure), the cross-type concurrency claim holds (separate stores/DBs), and the scope is minimal with no drive-by changes. Two nits:

- **Alert title casing**: `'Download in progress'` should be `'Download In Progress'` to match the file's other titles (`'Download Failed'`, `'Delete Translation'`).
- The title+body strings are duplicated across both handlers, so extract to a shared constant or a small `showDownloadInProgressAlert()` helper.

Optional: `lightHaptics()` fires before the guard, so blocked taps still buzz, which is fine as "acknowledge the tap," just worth a one-line comment so it reads as intentional.

omar-zarka:
Thanks both. Addressed osmansaeday's nits in `d9f0e21`:

- Title casing → `'Download In Progress'` (matches `'Download Failed'` / `'Delete Translation'`).
- Extracted the duplicated title+body into a shared `showDownloadInProgressAlert()` helper so both handlers stay in sync.
- Added a one-line comment on both handlers noting the pre-guard `lightHaptics()` is intentional (acknowledge the tap).

On the cross-guard (OKKHALIL3): that was already resolved in the earlier round (`a8c1b4c`) — `handleDownload` guards on `downloadingId` only and `handleDownloadTafseer` on `tafseerDownloadingId` only, with matching `useCallback` deps. I've also reframed the PR description so the severity reads as "make the refusal visible" rather than corruption-prevention (the store-level no-op already prevents that). `tsc` clean.

osmansaeday (approved):
Re-review: both nits addressed. The duplicated copy is extracted to a shared `showDownloadInProgressAlert()` helper, the title is now title-case (`'Download In Progress'`) matching the rest of the file, and the "haptics fire before the guard to acknowledge the tap" intent is documented in a comment. `useCallback` deps correctly include the own-type in-flight id. LGTM.

PR #296, opened 2026-06-01
chore(eslint): add explicit no-dupe-keys rule

omar-zarka:
## What

Adds an explicit `'no-dupe-keys': 'error'` to `.eslintrc.cjs`'s `rules` block.

## Why

`eslint:recommended` already enables `no-dupe-keys`, so this is **no behavior change today** — but declaring it explicitly:

- Makes the intent visible at the rules site (a maintainer scanning `rules` sees it's deliberately on).
- Survives any future reordering of `extends` (if `eslint:recommended` ever moves or is dropped, the guard stays).

It's cheap insurance against the JS last-key-wins footgun in large config objects. A fork of this repo (Qariah) hit exactly this: a duplicate top-level key in `config/branding.js` silently shadowed the intended value with the inherited default, and static analysis didn't flag it because the lint job hadn't run on that change. Making the rule explicit (and mirrored across forks) keeps it loud.

## Test

`eslint .` passes unchanged. Verified the rule fires on a synthetic duplicate-key object literal.


OKKHALIL3 (commented):
The 4-line comment works against itself, the second sentence ("eslint:recommended already enables this") undercuts the case for
 the rule, and the config/branding.ts callout references a future RFC that hasn't landed yet, which will read as stale noise the
 moment this merges. If you want to keep a comment at all, one line is enough: // Explicit so it survives any future reordering of
 'extends'. Otherwise just drop it, the rule name is self-documenting.

PR #298, opened 2026-06-01
chore(BrowseGrid): wire keyboardShouldPersistTaps prop through to LegendList

omar-zarka:
## What

`BrowseGridProps` declares `keyboardShouldPersistTaps?: 'always' | 'handled' | 'never'`, but the component never destructured it — the inner `LegendList` hardcoded `keyboardShouldPersistTaps="handled"`. So any caller passing a different value had it silently dropped (a dead prop).

This destructures the prop with a default of `'handled'` and passes it through to `LegendList`.

## Why

Follow-up to the review thread on #274 (option **(a)** — wire it through, vs (b) drop from the interface). Wiring it through keeps the declared API meaningful: a caller that needs a different keyboard-dismiss policy (e.g. a search list that wants tap-outside-to-dismiss) can now set it.

Byte-equivalent behavior for every current caller — the `= 'handled'` default reproduces the old hardcoded value.

## Test

`tsc --noEmit` clean; `prettier --check` clean. No caller change needed; existing callers keep the `'handled'` behavior.


osmansaeday (commented):
Correct fix. The prop was declared-but-undestructured (a dead prop), the `= 'handled'` default reproduces the old hardcoded value byte-for-byte, and `LegendList` accepts `keyboardShouldPersistTaps` natively via `ScrollViewProps`.

One latent footgun: the hand-rolled `React.memo` comparator (lines ~180-185) doesn't compare `keyboardShouldPersistTaps`, so a caller that changes it dynamically would be silently skipped, which is exactly the configurability this PR is enabling. Add it to the comparator (or reconsider whether that custom comparator earns its keep given how many props it already omits).

omar-zarka:
Addressed in `e00a6a7` — added `keyboardShouldPersistTaps` to the hand-rolled `React.memo` comparator so a dynamic change to the now-configurable prop participates in the re-render decision. Kept the custom comparator as-is to keep this focused; happy to file a follow-up to revisit whether it still earns its keep given the props it omits.

osmansaeday (approved):
Re-review: addressed. `keyboardShouldPersistTaps` is now included in the `React.memo` comparator, so a caller that changes it dynamically will correctly re-render. Default `'handled'` still reproduces the prior behavior for existing callers. LGTM.

PR #299, opened 2026-06-01
docs(rfc-018): community-reflections provider + render-slot seams

omar-zarka:
Doc-only RFC (per the doc-RFC-first convention) proposing two paired optional `Branding` slots so a fork can surface community-authored ayah reflections **without forking the hot verse-render path**.

## The seam

1. `communityReflectionsProvider?: (chapterId, verseNumber, locale?) => Promise<CommunityReflection[]>` — per-ayah data fetcher.
2. `ayahCommunityReflectionsComponent?: ComponentType<{chapterId, verseNumber}>` — RFC-008-style render slot mounted under the Arabic line, gated behind a new `MushafSettingsStore.showCommunityReflections` toggle.

Both unset (Bayaan default) → nothing renders, no toggle row appears, **zero behavior change**.

## Why

The Qariah fork ships this end-to-end in production (inline top-reflection under each ayah + a fullscreen popup from the verse action sheet, sourced from QuranReflect). To do it, Qariah diverges on `VerseItem.tsx` and `VerseActionsSheet.tsx` — two of the highest-churn files in the player — so every upstream merge re-conflicts. The *source* and *content* are fork-specific, but the two integration points ("fetch reflections for this ayah" / "render something under the Arabic line") are generic. This is the same divergence shape RFC-008 (Listen-tab slot) and RFC-009 (translation/tafsir providers) resolved.

Fifth entry in the XF-NNN multi-tenancy series. Full motivation, interface, alternatives (data-only / generic slot-array / baked-in source), and implementation plan in the doc. Code PRs (Bayaan ~120 LOC + Qariah-side) follow on shape approval.

Pre-flight: grepped `develop` for any existing community-content surface — none, so this is a clean additive seam.


osmansaeday:
### Code review

Found 1 issue:

1. The proposed seam contract uses `chapterId` / `verseNumber`, names that appear nowhere in the codebase; every existing verse-facing type and component uses `surahNumber` / `ayahNumber` (`VerseActionsSheet`, `types/timestamps.ts`, `types/tafseer.ts`, `types/verse-annotations.ts`). Since this is a fork-facing contract that Qariah will implement against, locking divergent names in at the RFC stage creates a rename-or-shim problem later. Suggest `surahNumber` / `ayahNumber` for both the provider signature and the slot props.

https://github.com/thebayaan/Bayaan/blob/203bbfb67362fb174ff3a444c7988d2538438210/docs/rfcs/018-community-reflections-provider.md#L49-L52

https://github.com/thebayaan/Bayaan/blob/203bbfb67362fb174ff3a444c7988d2538438210/docs/rfcs/018-community-reflections-provider.md#L81-L85

https://github.com/thebayaan/Bayaan/blob/85de88238a2e2e1665dc5b1b6d49431a2b27a229/components/sheets/VerseActionsSheet.tsx#L118-L121

omar-zarka:
Addressed in 653785d — renamed `chapterId`/`verseNumber` → `surahNumber`/`ayahNumber` across the provider signature, the `CommunityReflectionsProvider` type, the `ayahCommunityReflectionsComponent` slot props, and the prose/alternatives, to match the codebase convention (`VerseActionsSheet` / `types/timestamps.ts` / `types/tafseer.ts` / `types/verse-annotations.ts` all use `surahNumber`/`ayahNumber`). 6 + 6 occurrences updated; no other identifiers touched. Thanks for the catch.

osmansaeday (commented):
All seven factual claims verified: RFC-008 (component slot) and RFC-009 (provider) precedents match, `VerseItem`/`VerseActionsSheet` are the right integration points, `MushafSettingsContent` already imports `branding` so the gate pattern is established, and the "no existing community-content surface on develop" grep is accurate. The two-slot (data + render) shape vs RFC-009's provider-only is intentional and well-argued. Three things to tighten before the code PR:

1. **Provider<->component coupling is ambiguous (latent double-fetch).** The provider returns `Promise<CommunityReflection[]>` but the render slot only gets `{surahNumber, ayahNumber}`, so nothing hands the fetched data to the component, which makes the top-level `communityReflectionsProvider` field effectively just a gate-predicate while the component re-fetches. State explicitly that the component calls the provider internally (and caching is what keeps that from being two fetches), or pass the data/provider down as a prop.

2. **Blast radius.** `VerseItem` is rendered by **three** callers: the player's FlashList (`QuranView`), `ContinuousListView`, and `ReadingPageView`, not just "the player Mushaf list." A slot mounted unconditionally in `VerseItem` appears in all three. Either acknowledge that or scope it via the existing `source` prop.

3. **Hot-path contract.** This provider runs per-ayah on a recycling FlashList (unlike RFC-009's cold Settings flows). Specify that the slot must be non-blocking, tolerate rapid mount/unmount/key churn from recycling, and ideally sit behind an ErrorBoundary at the mount site (the same gap RFC-008 deferred, more acute here since this slot does I/O).

Approving the shape with these comments.

omar-zarka:
Tightened in `e4c830c` before the code PR — which is now open at #306:

1. **Coupling / double-fetch:** documented that the render slot calls `communityReflectionsProvider` **internally** (the top-level field is both gate-predicate and data source) and that provider-internal caching keyed by `(surah, ayah[, locale])` is load-bearing for keeping inline + popup to a single fetch; also wrote up the pass-the-data-down-as-a-prop alternative if you'd rather the host own the lifecycle.
2. **Blast radius:** acknowledged `VerseItem` renders in `QuranView`, `ContinuousListView`, and `ReadingPageView` — v1 mounts in all three, with the existing `source` prop documented as the narrowing knob.
3. **Hot-path contract:** non-blocking, recycling-tolerant (drop late responses for a now-stale ayah on prop churn), behind an ErrorBoundary at the mount site.

The code PR (#306) implements exactly this — a nullable seam that's a no-op for Bayaan.

osmansaeday (approved):
Re-review: all three comments from my earlier review are now documented in the RFC, and the implementation lands in #306.

- **Provider<->component coupling** — a "review reconciliation" block states the component calls the provider internally and that provider-internal caching is load-bearing (so the inline slot and the action-sheet popup share one fetch).
- **Blast radius** — names all three `VerseItem` callers (`QuranView`, `ContinuousListView`, `ReadingPageView`), with the `source` prop offered as the narrowing knob; v1 mounts in all three.
- **Hot-path contract** — non-blocking, recycling-tolerant (abort/ignore stale-ayah responses), and ErrorBoundary-at-mount-site recommended.

Shape approved. LGTM.

PR #300, opened 2026-06-06
perf(mushaf): dispose Skia paragraphs on recompute/unmount

omar-zarka:
`SkParagraph` objects hold native memory that JS GC never reclaims. The three Skia paragraph-building views in the Mushaf rebuild paragraphs inside `useMemo` as pages/verses recompute, and the old paragraphs were never freed — pinning the native heap and forcing explicit GCs (a Pixel 3 profiling pass measured ~164 MB native heap + repeated explicit GCs during reading).

This adds a `useEffect` cleanup keyed on the memoized paragraph(s) in `SkiaLine`, `SkiaVerseText`, and `WBWVerseView`. React runs the cleanup with the **previous** memo value when the dependency changes and on unmount, so the paragraphs being rendered the current frame are never disposed early. `.dispose()` is already the disposal mechanism used for the transient measurement paragraphs in `JustificationService`.

**Default preserves current behavior** — the rendered frame is identical; this only frees native memory that was previously leaked on recompute/unmount. Three focused hunks.

## Test plan & results

Verified on a **Pixel 3 (Android 12) release build** — a downstream fork build whose copies of these three files are byte-identical to `develop`, so it's a faithful proxy on the exact device where the heap was profiled. `tsc --noEmit` clean.

| Scenario (the use-after-free / blank-render risks) | Result |
|---|---|
| `SkiaVerseText` — 18 fast scroll churns through the verse list (memo recompute → dispose) | text renders correctly; **no native crash** |
| `SkiaPage`/`SkiaLine` — 24 page-turns fwd+back across Al-Baqarah | pages render crisply, no blank/missing glyphs; **no native crash** |
| Rapid Mushaf-settings switches (page-design / view-type / scroll) forcing remount+dispose | correct re-render; **no native crash** |
| Verse-actions sheet opened over the churned state | opens + renders fine |
| Full-session logcat | **zero** FATAL / SIGSEGV / SIGABRT / libskia-abort / use-after-free / JNI-DETECTED |

`WBWVerseView` uses the identical cleanup pattern and is tsc-clean; its display mode wasn't separately reached on-device in this pass (same mechanism → low incremental risk). Steady-state PSS is intentionally **not** quoted as an A/B: native allocators pool freed memory rather than return it to the OS, and the resident heap is dominated by the (intentionally retained) mushaf preload — so the win here is reduced GC pressure / leak-prevention per the profiling above, not a headline MB drop.

### DIFF
--- a/components/mushaf/skia/SkiaLine.tsx
+++ b/components/mushaf/skia/SkiaLine.tsx
@@ -50,6 +50,7 @@ interface SkiaLineProps {
     paragraph: SkParagraph,
     xPos: number,
   ) => void;
+  onParagraphDisposed?: (lineIndex: number) => void;
   backgroundHighlights?: Array<{start: number; end: number; color: string}>;
 }
 
@@ -69,6 +70,7 @@ const SkiaLine: React.FC<SkiaLineProps> = ({
   fontFamily = 'DigitalKhatt',
   arabicTextWeight = 'normal',
   onParagraphReady,
+  onParagraphDisposed,
   backgroundHighlights,
 }) => {
   const paragraphs = useMemo(() => {
@@ -209,12 +211,22 @@ const SkiaLine: React.FC<SkiaLineProps> = ({
   // previous paragraphs when the memo recomputes (and on unmount). The cleanup
   // runs with the prior `paragraphs` value, so the set being rendered this
   // frame is never disposed early.
+  //
+  // Evict the parent's hit-test map entry in the SAME synchronous cleanup,
+  // BEFORE disposing, so the map never points at freed native memory. The
+  // recompute-to-new-paragraph path re-registers via `onParagraphReady`
+  // immediately after this cleanup; the recompute-to-null and unmount paths
+  // leave the entry evicted (the cases the parent's pageNumber-only clear
+  // misses).
   useEffect(() => {
     return () => {
+      if (paragraphs) {
+        onParagraphDisposed?.(lineIndex);
+      }
       paragraphs?.paragraph?.dispose();
       paragraphs?.strokeParagraph?.dispose();
     };
-  }, [paragraphs]);
+  }, [paragraphs, lineIndex, onParagraphDisposed]);
 
   const lineInfo = quranTextService.getLineInfo(pageNumber, lineIndex);
   const maxWidth = pageWidth * 2;
--- a/components/mushaf/skia/SkiaLine.tsx
+++ b/components/mushaf/skia/SkiaLine.tsx
@@ -205,6 +205,17 @@ const SkiaLine: React.FC<SkiaLineProps> = ({
   const paragraph = paragraphs?.paragraph;
   const strokeParagraph = paragraphs?.strokeParagraph;
 
+  // SkParagraph holds native memory that JS GC does not reclaim. Dispose the
+  // previous paragraphs when the memo recomputes (and on unmount). The cleanup
+  // runs with the prior `paragraphs` value, so the set being rendered this
+  // frame is never disposed early.
+  useEffect(() => {
+    return () => {
+      paragraphs?.paragraph?.dispose();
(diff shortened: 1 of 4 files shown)


osmansaeday (changes requested):
Solid direction: disposing `SkParagraph` native memory on recompute/unmount is the right fix, the cleanup-runs-with-prior-value reasoning is correct, and the null-guarding is careful. One memory-safety concern before merge.

**Use-after-free via `SkiaPage.paragraphMapRef` (should-fix).** `SkiaPage` registers each line's paragraph in `paragraphMapRef` (`SkiaPage.tsx:364`) and reads it during touch hit-testing (`getGlyphPositionAtCoordinate`, `SkiaPage.tsx:435`). That map is only invalidated on `pageNumber` change (`SkiaPage.tsx:344`). The new dispose effect in `SkiaLine` frees a paragraph in two paths where `onParagraphReady` does **not** re-fire to update the map:

- the memo recomputes to `null` (`SkiaLine.tsx:80`, `!lineText`), and
- a line unmounts while the page stays mounted.

In both, `paragraphMapRef[lineIndex]` keeps a now-disposed pointer, and the next hit-test reads freed native memory. The common case (recompute to a *new* valid paragraph) is safe, because `onParagraphReady` updates the map in the same synchronous effect flush, which is why the on-device testing didn't surface it. Suggested fix: have `SkiaLine` notify the parent to evict its entry before disposal (an `onParagraphDisposed(lineIndex)` callback, or call `onParagraphReady(lineIndex, null)`), so map invalidation is tied to disposal rather than only to `pageNumber`.

**Pre-existing leak worth fixing while you're here (should-fix).** `WBWVerseView.measureTextWidth` builds a `SkParagraph` for measurement and returns a number without `.dispose()`, so it leaks on every recompute and now sits right next to the disposal infra this PR adds.

Nits: the dep arrays and null-guards all look correct; the `built` rename in `SkiaVerseText` to keep the whole memo object for the effect dep is a clean choice.

omar-zarka:
Thanks — both fixed in `ad0562e`, with map invalidation tied to disposal as suggested:

- **Use-after-free via `paragraphMapRef`.** `SkiaLine` now takes an `onParagraphDisposed(lineIndex)` prop and calls it inside the existing dispose-effect cleanup **before** `dispose()`; `SkiaPage` passes a stable callback that deletes the map entry. Walking the paths: recompute→new re-registers via `onParagraphReady` in the same commit (the synchronous-flush case you noted is safe); recompute→null and unmount leave the entry evicted. The hit-test already null-guards the map read, so an evicted entry is a safe no-op. I used a dedicated callback rather than `onParagraphReady(i, null)` since the latter's signature is non-nullable (faking it would need a cast the lint disallows).
- **`measureTextWidth` leak.** Now disposes the measurement paragraph via `try/finally` on every return path.

`tsc` clean; only the three files touched.

osmansaeday (approved):
Re-review: both findings from my earlier review are resolved.

- **Use-after-free (was blocking) — RESOLVED.** `SkiaLine` now takes an `onParagraphDisposed(lineIndex)` callback and fires it from the dispose effect cleanup, synchronously before `.dispose()`, and `SkiaPage.handleParagraphDisposed` deletes the `paragraphMapRef` entry. Traced all three paths (recompute-to-null, recompute-to-new, unmount): the map never holds a disposed pointer that hit-testing can read. Eviction only shrinks the map, so the `readyFiredRef`/`onReady` threshold can't mis-fire.
- **`measureTextWidth` leak (was should-fix) — RESOLVED.** Wrapped in try/finally with `dispose()` after `getLongestLine()`, so it frees on every path. The extra `WBWVerseView` effect also disposes the full `computedLayout` paragraph set on recompute/unmount.

Clean. The evict-before-dispose ordering is documented at the cleanup site, which is exactly the invariant a future reader needs. LGTM.

PR #301, opened 2026-06-06
perf(mushaf): bound and clear the page-layout cache

omar-zarka:
`pageLayoutsCache` in `JustificationService` grew without limit and was never cleared. It partitions by `font::rewayah::ratio`, and each partition can hold up to 604 page layouts, so switching rewayah / font / size left stale partitions resident for the whole session.

- Cap each partition's page map at the **50 most-recently-used** pages, evicting the oldest on overflow (Map insertion order; re-inserting on hit keeps recent pages live). The MMKV layer still backs evicted pages, so recent-page hits are unaffected.
- Add `JustService.clearPageLayoutCache()` and call it from `QuranTextService.clearCaches()` — which already runs on rewayah change — so stale partitions are dropped on exactly the transition that invalidates them (lazy `require` to avoid the JustificationService ⇄ QuranTextService import cycle, matching the existing pattern in both files).

**Default preserves current behavior** for the common case: recent pages still hit in memory; this only bounds growth and clears layouts that were already being recomputed after a rewayah/font change.

## Test plan & results

Verified on a **Pixel 3 release build** (files byte-identical to `develop`). `tsc --noEmit` clean.

| Scenario (the dropped-needed-layout / wrong-page risks) | Result |
|---|---|
| Repeatedly toggle PAGE DESIGN (Fullscreen↔Book) and VIEW TYPE (Mushaf↔List) — each switches the layout partition and runs `clearCaches` | page re-renders correctly every time; **no blank / wrong page**; no crash |
| 24 page-turns to populate the cache, then leave the Mushaf (triggers `clearCaches`) | correct render throughout; Graphics PSS released on leave (206 → 132 MB) |
| Full-session logcat | no FATAL / native abort |

Two files, tsc-clean.

### DIFF
--- a/services/mushaf/JustificationService.ts
+++ b/services/mushaf/JustificationService.ts
@@ -929,8 +929,13 @@ export class JustService {
         fontFamily,
       );
       if (mmkvHit) {
-        // Populate in-memory cache for fastest subsequent access
-        pageCache.set(pageNumber, mmkvHit);
+        // Populate in-memory cache for fastest subsequent access (bounded)
+        JustService.cachePageLayout(
+          fontSizeLineWidthRatio,
+          pageNumber,
+          mmkvHit,
+          fontFamily,
+        );
         return mmkvHit;
       }
     } catch {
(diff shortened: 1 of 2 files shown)


osmansaeday (commented):
Verified: the bound is correct and safe. `cachePageLayout` does `delete` -> evict-oldest-on-overflow -> `set` (cap holds at 50, no off-by-one), the MMKV-rewarm path now routes through the bounded setter, and `clearPageLayoutCache` is wired into `clearCaches` with the lazy-require cycle break. Correctness is preserved because every eviction is backed by MMKV.

One accuracy note (non-blocking): the "most-recently-used / re-inserting on hit keeps recent pages live" framing only holds for the write/rewarm path. The in-memory read hit (`getCachedPageLayout`, lines 921-922) returns directly without re-inserting, so a page that stays resident and is re-read never has its recency bumped, which makes eviction insertion/rewarm-order rather than access-LRU. If you want true MRU, add a `delete`+`set` on the memory hit; otherwise consider softening the wording. Either way it's a marginal (~1ms MMKV re-read) effect, not a correctness issue.

Nit: `MAX_CACHED_PAGES_PER_PARTITION` is declared below its first use (works via runtime ordering, but reads oddly), consider hoisting it with the other module constants. The partition *count* itself stays unbounded (one per font-size ratio dragged), worth a follow-up but benign.

omar-zarka:
Both addressed in `65dae2c`:

- Hoisted `MAX_CACHED_PAGES_PER_PARTITION` up to the module-constant block (now declared before its first use).
- Softened the wording rather than changing behavior: the comment now describes eviction as insertion/rewarm order and calls out that an in-memory read hit doesn't re-insert (so a resident re-read keeps its original recency — not strict access-LRU), with the MMKV backing as the ~1ms safety net. Also noted the partition *count* itself stays unbounded (one per font-size ratio) as a benign follow-up.

osmansaeday (approved):
Re-review: feedback addressed. The MRU-vs-FIFO wording is now honest — the comment documents that eviction is insertion/rewarm order (not strict access-LRU), that a resident page re-read keeps its original recency, and that every eviction is MMKV-backed (~1ms re-read). `MAX_CACHED_PAGES_PER_PARTITION` is hoisted to the top with the other module constants, and the unbounded-partition-count caveat is noted for a follow-up. LGTM.

PR #302, opened 2026-06-06
docs(rfc): RFC-019 bundled timestamp provider seam (timestampLocalProvider)

omar-zarka:
Doc-first RFC proposing an optional `branding.timestampLocalProvider?: (rewayatId, surahNumber) => AyahTimestamp[] | null` so a fork can supply **bundled** ayah-timing data (authored offline, baked into the app) instead of serving it from a CDN.

Today the only way to wire bundled timings into `TimestampFetchService` is to shadow the whole file — which gets dropped on every wholesale upstream absorption of that file (RFC-015's `R2_BASE` rework and PR #278's rewrite both touched it). This is the bundled-data counterpart to RFC-015's read-side CDN seam, with the same inline-fallback shape: the consumer applies `?? null` at each of the three resolution methods (`hasSource` / `hasSurah` / `fetchAndCache`), so **Bayaan's behavior is byte-identical when the field is unset**. The `'local'` `TimestampSource` it targets already exists in the upstream schema, which strengthens the "why now".

Numbered **019** (017 = server-driven config envelope #291, 018 = community-reflections provider #299 are already claimed).

**Doc-only** — there's no runtime to test; the verification for this PR is the shape review. The main thing I'd like confirmed is the `hasSource` presence-probe shape (see Open Questions). The consumer touches the timestamp fetch path, so per the unusual-shape convention this lands doc-first; the **code PR follows on shape approval**.

### DIFF
--- a/docs/rfcs/019-timestamp-local-provider.md
+++ b/docs/rfcs/019-timestamp-local-provider.md
@@ -1,4 +1,4 @@
-# RFC-019: Bundled timestamp provider seam (`branding.timestampLocalProvider`)
+# RFC-019: Bundled timestamp provider + coverage-list seam pair (`branding.timestampLocalProvider` + `timestampLocalSurahList`)
 
 **Status:** Draft (doc-first — the consumer touches the timestamp fetch
 hot-path in three methods, so the shape is worth confirming before code lands;
@@ -30,10 +30,13 @@ This RFC adds an optional **provider pair**:
   — returns the bundled timings for one surah, or `null`.
 - `branding.timestampLocalSurahList?: (rewayatId) => number[] | null` — the
   **explicit coverage signal**: the list of surahs the bundle covers for a
-  rewayat (`null` / empty → no local coverage). This mirrors the catalog's
-  existing `Rewayat.timestamps_surah_list?: number[]` exactly, so `hasSource` /
+  rewayat (`null` / empty → no local coverage). This mirrors the *shape* of the
+  catalog's existing `Rewayat.timestamps_surah_list?: number[]`, so `hasSource` /
   `hasSurah` can answer coverage questions **without** calling the data
-  provider or guessing from a sentinel surah.
+  provider or guessing from a sentinel surah. One semantic difference from the
+  R2 rule — the local list is an **exact allow-list** (empty means *no*
+  coverage, not "all"); see [Coverage
+  semantics](#coverage-semantics-local-is-an-exact-allow-list-not-empty-means-all).
 
 When set, the three consumer methods consult the pair before going to the
 network — `hasSource` / `hasSurah` gate on the coverage list, `fetchAndCache`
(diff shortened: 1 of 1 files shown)


osmansaeday:
### Code review

Found 2 issues:

1. The `hasSource` surah-1 presence probe breaks the RFC's own core use case: partial-coverage reciters. A fork bundling timings for a reciter that lacks surah 1 (28 of ~268 rewayat in the current catalog have no surah 1 in `surah_list`) gets `null` from the probe, `hasSource` returns false, and the follow-along feature is suppressed entirely despite valid local data, since `TimestampService.hasTimestampSource` gates the UI on it. Open question 1 frames the probe shape as a taste question, but it is a correctness hole; the provider needs an explicit coverage signal (your alternative (b), a surah list next to the provider) rather than a sentinel probe.

https://github.com/thebayaan/Bayaan/blob/0efec5ab6ac7e8bd5440cd3dc95649f1b1f6fb10/docs/rfcs/019-timestamp-local-provider.md#L124-L138

https://github.com/thebayaan/Bayaan/blob/0efec5ab6ac7e8bd5440cd3dc95649f1b1f6fb10/docs/rfcs/019-timestamp-local-provider.md#L191-L199

2. The `fetchAndCache` sketch writes the provider's array straight to `writeTimestamps(..., 'local')` with no shape validation, bypassing the `isAyahTimestampShape` guard the R2 path gained after the RFC-015 review (PR #286). The threat model is identical (fork-authored data; malformed entries land `undefined` in the SQLite `duration_ms` column and produce NaN highlight offsets), so the local path should run the same first-element probe before writing.

https://github.com/thebayaan/Bayaan/blob/0efec5ab6ac7e8bd5440cd3dc95649f1b1f6fb10/docs/rfcs/019-timestamp-local-provider.md#L150-L164

https://github.com/thebayaan/Bayaan/blob/85de88238a2e2e1665dc5b1b6d49431a2b27a229/services/timestamps/TimestampFetchService.ts#L84-L93

omar-zarka:
Thanks Osman — both are real holes, fixed in `3c50492` (doc-only).

**Issue 1 — partial-coverage suppression (correctness, not taste).** You're right that the surah-1 sentinel probe breaks the RFC's own core use case. I've replaced it with the explicit coverage signal — the draft's own alternative (b): a companion `branding.timestampLocalSurahList?: (rewayatId) => number[] | null` that mirrors the catalog's existing `Rewayat.timestamps_surah_list?: number[]` one-to-one.

- `hasSource` / `hasSurah` now gate on the coverage list (`localSurahs(rewayatId).length > 0` / `.includes(surahNumber)`) — never the data provider, never a sentinel surah. A reciter whose bundle lacks surah 1 just lists the surahs it has, and the follow-along UI is enabled for exactly those.
- The data provider (`timestampLocalProvider`) is now called *only* on the `fetchAndCache` write path.
- Open question 1 is resolved in-doc accordingly (re-marked RESOLVED with the 28/268 partial-coverage rationale, not left as an open shape question).

**Issue 2 — local writes bypassed the `isAyahTimestampShape` guard.** Agreed — identical threat model to the R2 path (fork-authored data of untrusted shape → malformed entry lands `undefined` in `duration_ms` → NaN highlight offsets). The `fetchAndCache` sketch now runs the same first-element probe (the guard added in #286) before `writeTimestamps(..., 'local')`:

```diff
+      if (
+        Array.isArray(local) &&
+        local.length > 0 &&
+        isAyahTimestampShape(local[0])
+      ) {
+        await timestampDatabaseService.writeTimestamps(rewayatId, surahNumber, local, 'local');
+        return local;
+      }
+      // coverage claimed but data missing/malformed → fall through to R2
+      // rather than caching garbage
```

It reuses the existing module-level guard verbatim — no new validator. Precedence (OQ-3) is now stated precisely: local wins when the coverage list claims the surah **and** the provider returns valid data; otherwise it falls through to R2.

Net upstream diff is now two branding types + one private `localSurahs` helper + three guarded early-outs; `config/branding.js` stays zero-diff. Ready for another look.

osmansaeday (commented):
Unusually well-researched RFC. I verified every codebase claim and they all hold: the three resolution methods (`hasSource`/`hasSurah`/`fetchAndCache`) exist as named, `'local'` is already in the `TimestampSource` schema, `writeTimestamps` accepts it, `AyahTimestamp` matches the proposed shape, the `isAyahTimestampShape` reuse is faithful, and RFC-015's `?? base` fallback shape is the right symmetry. Numbering (017/018/019) is collision-free. The sync-provider decision is sound (no real sync/async impedance at the three call sites), and resolving the `hasSource` open question with a `timestampLocalSurahList` companion (mirroring `timestamps_surah_list`) is the correct call.

One thing to document before the code PR: the proposed `hasSurah` gate is an **exact allow-list** (`.includes(surah)`), whereas the existing R2 path treats `has_timestamps && empty surah_list` as "all surahs covered." That asymmetry is fine and arguably better, but it's currently presented as symmetric, so please call it out explicitly so the consumer PR doesn't reintroduce the "empty => all" rule by accident, and note that local coverage bypasses `has_timestamps`.

Nit: the title says `timestampLocalProvider` (singular) but the design is really a provider + coverage-list pair. Approving the shape.

omar-zarka:
Documented in `7377ea7` before the code PR — which is now open at #305:

- Added a "Coverage semantics" section: the local `timestampLocalSurahList` is an **exact allow-list** (`.includes(surah)`; empty/`null` = *no* coverage, no "everything" sentinel), whereas the R2 path keeps `has_timestamps && empty surah_list` = "all surahs covered." The two read an empty list oppositely, with an explicit "**do not** copy the R2 'empty ⇒ all' branch onto the local path" warning so the consumer PR can't reintroduce it by reflex.
- Documented that **local coverage bypasses `has_timestamps`** (the local checks short-circuit before the flag read).
- Retitled to reflect it's a provider + coverage-list **pair**.

The code PR (#305) implements the exact-allow-list + `has_timestamps`-bypass semantics; `config/branding.js` is unchanged so Bayaan is a no-op.

osmansaeday (approved):
Re-review: the two semantic deltas I asked to document are now in the RFC, and the implementation lands in #305.

- A "Coverage semantics: local is an exact allow-list, not empty-means-all" section spells out that empty-on-R2 means "all" while empty-on-local means "none", with an explicit "Do not copy the R2 empty => all branch onto the local path" warning.
- The `has_timestamps` bypass for local coverage is documented (the local checks sit above the `findRewayat`/`has_timestamps` lines).

Shape approved. LGTM.

PR #303, opened 2026-06-06
feat(sentry): tag release/dist + breadcrumb cold-start for crash observability

omar-zarka:
Two small, generic observability changes in `app/_layout.tsx` — both no-config-change and safe.

## 1. Explicit `release` + `dist` in `Sentry.init`

Android crash/ANR events were landing in Sentry with `release: None` — Sentry's native release auto-detection isn't reliable on Android, so an event couldn't be tied to a build. This derives `release` + `dist` from the build-time version that `scripts/generate-version.js` already injects into `expoConfig.extra.version`, and sets them explicitly in `Sentry.init`.

```ts
release: `${appId}@${semanticVersion}+${buildNumber}`
dist: String(buildNumber)
```

If the manifest is somehow missing the version info, both resolve to `undefined`, which is a no-op (Sentry just falls back to its existing auto-detect) — so this can't regress current behavior. `appId` reads `ios.bundleIdentifier ?? android.package`, no hardcoded fallback.

## 2. Cold-start breadcrumbs + slow-boot watchdog

A hang at the splash screen is otherwise invisible in Sentry — nothing breadcrumbs the boot path, so a "stuck on logo" report has no trail. This adds:

- a `markBoot(step)` helper that drops a `boot` breadcrumb at each real phase in `prepare()` (audio init → catalog fetch → app initializer → session restore → ready), so any later crash carries the boot trail, and
- an 8s `setTimeout` watchdog that fires `captureMessage('slow-cold-start', { tags: { boot_step } })` if init blows past budget, `clearTimeout`'d in a `finally`. This turns the invisible boot hang into a queryable signal that names the phase it stalled in.

## Notes

- No new dependency — `expo-constants` is already a transitive dep; just added the import.
- Verified with `tsc --noEmit` (clean).
- Diff is +50 / −0, single file.

### DIFF
--- a/app/_layout.tsx
+++ b/app/_layout.tsx
@@ -61,6 +61,11 @@ configureReanimatedLogger({
 // Cache for expo-navigation-bar module (Android only)
 let NavigationBarModule: any = null;
 
+// Cold-start budget: if prepare() blows past this, fire a one-shot
+// 'slow-cold-start' Sentry message naming the phase it stalled in. prepare()
+// is gated by initializationRef, so this fires at most once per process.
+const SLOW_BOOT_THRESHOLD_MS = 8000;
+
 // Prevent the splash screen from auto-hiding
 SplashScreen.preventAutoHideAsync().catch(() => {
   /* reloading the app might trigger some race conditions, ignore them */
@@ -74,6 +79,18 @@ SystemUI.setBackgroundColorAsync(
 
 const analyticsEnabled = process.env.EXPO_PUBLIC_ANALYTICS_ENABLED !== 'false';
 
+// Narrows the `any`-typed expoConfig.extra.version without an `as` cast.
+function isVersionInfo(
+  value: unknown,
+): value is {semanticVersion: string; buildNumber: string | number} {
+  if (value == null || typeof value !== 'object') return false;
+  const v = value as Record<string, unknown>;
+  return (
+    typeof v.semanticVersion === 'string' &&
+    (typeof v.buildNumber === 'string' || typeof v.buildNumber === 'number')
+  );
+}
+
 if (analyticsEnabled) {
   // Explicit release + dist tagging. Android events were landing with no
   // release (`release: None`), so crashes/ANRs couldn't be tied to a build.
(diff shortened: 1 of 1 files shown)


osmansaeday (commented):
Good observability win, and verified accurate: the `semanticVersion`/`buildNumber` field names match `generate-version.js`, `expo-constants` is a direct dep, and `clearTimeout` in `finally` covers every path through `prepare()`. Two should-fixes:

**`as` cast (should-fix, project convention).** `Constants.expoConfig?.extra?.version as {...}` uses an `as` cast, which the repo's TS rules avoid. `extra` is already `any`, so a small type guard (`function isVersionInfo(v: unknown): v is {...}`) narrows it without the cast.

**`release` can become the literal `"undefined@2.2.1+935"` (should-fix).** The guard checks `semanticVersion`/`buildNumber` but not `appId`. If `ios.bundleIdentifier`/`android.package` are absent at runtime, the template literal stringifies `undefined` into the release name rather than falling back to `undefined`. Bayaan always sets these, but your fork is the consumer most likely to hit it. Add `appId != null` to the condition.

Nit: consider a named `SLOW_BOOT_THRESHOLD_MS` constant. 8s can false-positive on first-install/slow-network cold starts where catalog fetch + initialize legitimately stack, and there's no rate-limit on the `captureMessage`.

omar-zarka:
Thanks — all three addressed in `bd3b127`:

- **`as` cast → type guard.** Replaced `Constants.expoConfig?.extra?.version as {…}` with an `isVersionInfo(value): value is {semanticVersion: string; buildNumber: string | number}` guard and narrow via it. `appId` isn't part of the guard — it's derived separately from `ios.bundleIdentifier ?? android.package`.
- **`undefined@…` release.** The condition now requires `versionInfo != null && appId != null`, so a missing bundle id falls back to an `undefined` release (Sentry auto-detects) instead of stringifying `undefined`.
- **Threshold constant.** Extracted `SLOW_BOOT_THRESHOLD_MS`. Didn't add a separate once-guard since `prepare()` is already gated by `initializationRef` (fires at most once per process) — noted that on the constant.

`tsc --noEmit` clean; only `app/_layout.tsx` touched.

osmansaeday (approved):
Re-review: both should-fixes from my earlier review are resolved.

- **`as` cast — RESOLVED.** The module-level `as` cast is replaced by an `isVersionInfo` type guard. The one remaining internal `as Record<string, unknown>` is the idiomatic, locally-contained narrowing of an already-checked `unknown` object — genuinely unavoidable, and the guard's predicate is actually stricter (required fields) than the old cast.
- **`release` => "undefined@..." — RESOLVED.** `versionInfo != null && appId != null` now gates the template literal, so a missing `appId` yields `undefined` (a Sentry no-op) rather than a stringified `"undefined@..."`.
- **Threshold — ADDRESSED.** `SLOW_BOOT_THRESHOLD_MS` is now a named constant, and the inline note documents that `initializationRef` makes `prepare()` (and thus the watchdog) fire at most once per process, which covers the rate-limit concern.

LGTM.

PR #304, opened 2026-06-07
feat(settings): add an analytics opt-out toggle (Settings → Privacy)

omar-zarka:
## What

Adds a user-facing control to turn off anonymous product-usage analytics, in **Settings → Privacy → "Share anonymous usage data."**

Today analytics can only be disabled at build time (`EXPO_PUBLIC_ANALYTICS_ENABLED`); there's no in-app way for a user to opt out. This adds one.

## How

- **`store/analyticsConsentStore.ts`** (new) — persisted consent flag (zustand + AsyncStorage), opt-out model, defaults to `true`. Mirrors the existing `devSettingsStore` shape.
- **`services/analytics/AnalyticsService.ts`**
  - `capture()` early-returns when the user has opted out.
  - new `applyConsent(enabled)` calls PostHog `optIn()` / `optOut()` (the SDK persists this across launches).
  - `setPostHogInstance()` applies the stored choice to the fresh instance.
- **`app/(tabs)/(d.settings)/index.tsx`** — a "PRIVACY" section with a toggle wired to the store + `applyConsent`.

## Notes

- No behavior change unless the user opts out (default on).
- Generic — no fork-specific strings (description uses `branding.appName`).
- `tsc --noEmit` + `prettier --check` clean on the three files. (This mirrors a change already shipped and verified in a downstream fork.)

## Test plan

- Settings → Privacy shows the toggle (default ON).
- Toggle OFF → `analyticsService.capture(...)` is suppressed and PostHog `optOut()` fires; choice survives an app relaunch.
- Toggle back ON → analytics resume.

### DIFF
--- a/__mocks__/@react-native-async-storage/async-storage.js
+++ b/__mocks__/@react-native-async-storage/async-storage.js
@@ -1,0 +1,6 @@
+// Jest automatically uses node-module mocks placed under a root-level __mocks__
+// directory. This lets test files import modules that transitively pull in
+// AsyncStorage (e.g. persisted zustand stores like analyticsConsentStore /
+// devSettingsStore) without the "[@RNC/AsyncStorage]: NativeModule:
+// AsyncStorage is null" error. Re-exports the package's official jest mock.
+module.exports = require('@react-native-async-storage/async-storage/jest/async-storage-mock');
--- a/app/(tabs)/(d.settings)/index.tsx
+++ b/app/(tabs)/(d.settings)/index.tsx
@@ -28,6 +28,8 @@ import {
   ShieldLockIcon,
 } from '@/components/Icons';
 import {useDevSettingsStore} from '@/store/devSettingsStore';
+import {useAnalyticsConsentStore} from '@/store/analyticsConsentStore';
+import {analyticsService} from '@/services/analytics/AnalyticsService';
 import {ThemePicker} from '@/components/settings/ThemePicker';
 import branding from '@/config/branding';
 
@@ -220,6 +222,12 @@ export default function SettingsScreen() {
   const styles = useMemo(() => createStyles(theme), [theme]);
   const glassColorScheme = useGlassColorScheme();
   const {showFloatingDevMenu, toggleFloatingDevMenu} = useDevSettingsStore();
+  const {analyticsEnabled, setAnalyticsEnabled} = useAnalyticsConsentStore();
+
+  const handleAnalyticsToggle = (value: boolean) => {
+    setAnalyticsEnabled(value);
+    analyticsService.applyConsent(value);
+  };
 
   const iconColor = theme.colors.text;
   const chevronColor = theme.colors.textSecondary;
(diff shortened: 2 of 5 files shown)


osmansaeday (changes requested):
Clean store that mirrors `devSettingsStore`, correct use of `getState()` (not the hook) inside the service, and `branding.appName` in the copy is right for the fork. Two data-handling issues before merge.

**Hydration race can send events for opted-out users (should-fix).** `zustand/persist` over `AsyncStorage` hydrates asynchronously. `analyticsConsentStore` isn't in the store pre-warm list, so on a cold launch `setPostHogInstance` -> `getState().analyticsEnabled` and the first `capture()` calls can both read the in-memory default (`true`) before the persisted `false` finishes loading, emitting events the user opted out of last session. Fix: gate on hydration, e.g. `await useAnalyticsConsentStore.persist.rehydrate()` in the initializer before `setPostHogInstance`, or apply consent from an `onFinishHydration` callback.

**`identifyUser` is ungated (should-fix).** `capture()` early-returns on opt-out, but `identifyUser` still calls `posthog.identify()`, which creates a person profile server-side (arguably more sensitive than a generic event). Add the same `if (!useAnalyticsConsentStore.getState().analyticsEnabled) return;` guard.

**Tests (should-fix).** The PR adds `optIn`/`optOut` to the test mock but no test actually asserts suppression, or that `applyConsent`/`setPostHogInstance` call `optOut()`/`optIn()`. Worth a couple of behavioral tests for the new suppression path (distinct from the existing env-flag tests).

Nits: `void (... optOut())` swallows rejections (at least `.catch(log)` since this is privacy-sensitive); consider wiring `applyConsent` into the store action so callers can't desync the SDK from the flag (would also drop the `analyticsService` import from the settings screen); missing explicit `: void` on `handleAnalyticsToggle`.

omar-zarka:
Thanks for the careful review — all four addressed in `5e90fdd`:

- **Hydration race.** New `isConsentGranted()` fails **closed** until `useAnalyticsConsentStore.persist.hasHydrated()`; both `capture()` and `identifyUser()` go through it, so nothing emits before the persisted choice loads. `setPostHogInstance()` applies consent immediately when hydrated, otherwise defers to `persist.onFinishHydration(...)` (matching the existing `waitForSettingsHydration` pattern in `MushafPreloadService`). I went with `onFinishHydration` over `await rehydrate()` so the call site stays synchronous inside the effect — the fail-closed gate guarantees correctness regardless.
- **`identifyUser` ungated.** Now guarded by the same `isConsentGranted()` (person profiles are more sensitive than generic events).
- **Tests.** Added five behavioral tests — `capture`/`identify` suppressed when opted out and resume on re-enable; `applyConsent`/`setPostHogInstance` drive `optOut()`/`optIn()`. The fake `optIn`/`optOut` now return `Promise<void>`.
- **Nits.** `applyConsent` logs rejections (`.catch`); wired `applyConsent` in via a consent-store subscription in the constructor so the SDK can't desync from the flag — the settings screen now only flips `setAnalyticsEnabled` and dropped its `analyticsService` import; explicit `: void` on `handleAnalyticsToggle`.

`tsc` clean; `AnalyticsService.test.ts` green.

osmansaeday (approved):
Re-review: all three findings from my earlier review are resolved.

- **Hydration race (was blocking) — RESOLVED.** New `isConsentGranted()` fails closed when `persist.hasHydrated()` is false, and gates both `capture()` and `identifyUser()`. `setPostHogInstance` applies the stored choice immediately if hydrated, else defers via `onFinishHydration` (with `unsub()` cleanup). The early-event window for opted-out users is closed.
- **`identifyUser` ungated (was blocking) — RESOLVED.** Same `isConsentGranted()` guard applied.
- **No behavioral tests (was should-fix) — RESOLVED.** Real suites now assert opted-out suppression of capture+identify, the applyConsent<->optOut/optIn linkage both directions, and the no-instance no-op.

Nits from last time (`.catch` on optIn/optOut, store-`subscribe` wiring so callers can't desync, explicit `: void`) all addressed.

One minor new note, not blocking: `setPostHogInstance` isn't idempotent — if the `posthog` ref ever changes identity it would register a second `onFinishHydration`/store subscription without cancelling the first. Not reachable with the current single-provider singleton setup; worth a stored-unsub guard if that ever changes. LGTM.

PR #305, opened 2026-06-19
feat(rfc-019): bundled local timestamp provider + coverage-list seam

omar-zarka:
Implements the code for **RFC-019** (#302) — shape approved ("Approving the shape"). Adds an optional, nullable multi-tenant seam so a fork can supply **bundled** (offline-authored) ayah timestamps for specific surahs, resolved entirely without R2/network — the bundled-data counterpart to RFC-015's read-side `timestampCdnBase`.

The upstream schema already carries the destination (`type TimestampSource = 'r2' | 'local'`; `writeTimestamps(…, 'local')`); the missing piece was a supported way for a fork to feed locally-authored timings into the resolver without shadowing `TimestampFetchService.ts`.

## API (`config/branding.d.ts` only — `config/branding.js` unchanged)
- `timestampLocalProvider?: (rewayatId, surahNumber) => AyahTimestamp[] | null` — bundled timings for one surah; consulted only on the data path.
- `timestampLocalSurahList?: (rewayatId) => number[] | null` — the explicit per-rewayat coverage signal, mirroring the catalog's `Rewayat.timestamps_surah_list`.

## Behavior (`TimestampFetchService` — the only code diff)
- `hasSource`/`hasSurah` gate on the **coverage list**, never the data provider and never a sentinel-surah probe (the RFC's resolved OQ1 — a surah-1 probe would falsely suppress follow-along for partial-coverage reciters whose bundle omits surah 1).
- `hasSurah` membership is an **exact allow-list** (`includes(surahNumber)`). Unlike the R2 path, an empty/unset local list means **no local coverage** — the R2 "empty `surah_list` ⇒ all" rule is deliberately not reintroduced for the local path.
- Local coverage **bypasses `has_timestamps`**.
- `fetchAndCache` runs the existing `isAyahTimestampShape` guard on the fork-supplied data, writes source `'local'` on success (subsequent reads short-circuit at the SQLite cache), and falls through to R2 on a missing/malformed bundle. Local wins over R2 (OQ3).

## No-op when unset (Bayaan default)
Both fields absent ⇒ the `?? null` / `?? []` coalesces short-circuit and every path is byte-equivalent to today; `config/branding.js` has zero diff.

## Verification
`npx tsc --noEmit` clean. Diff: 2 files (branding types + `TimestampFetchService`).

Related: RFC-015 (#286), RFC-009/RFC-018 (nullable `branding` provider precedent), PR #278 (introduced the fetch service).

### DIFF
--- a/config/branding.d.ts
+++ b/config/branding.d.ts
@@ -2,6 +2,7 @@ import type {ComponentType} from 'react';
 import type {Track} from '@/types/audio';
 import type {TranslationProvider} from '@/types/TranslationProvider';
 import type {TafsirProvider} from '@/types/TafsirProvider';
+import type {AyahTimestamp} from '@/types/timestamps';
 
 /** Catalog source config for the active branding. */
 export interface BrandingCatalogConfig {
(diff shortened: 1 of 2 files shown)


osmansaeday (commented):
Both semantic deltas from the doc review are correctly implemented: local coverage is an exact allow-list (`localSurahs(rewayatId) ?? []`, empty => no coverage, the inverse of R2's empty => all), and it bypasses `has_timestamps` (the local check precedes `findRewayat`/`has_timestamps` in both `hasSource` and `hasSurah`). Resolution order is right (local wins, falls through to R2 on null/malformed without stranding the R2 path or double-writing), `hasSource` gates on the coverage list and never calls the data provider (correct OQ1 resolution), and byte-equivalence when both fields are unset is confirmed (`?? []`/`?.()` short-circuits, no `as`/`any`, no import cycle).

**Should-fix (defensive, fork-facing):** the data path is guarded but the coverage path isn't. `TimestampService.getTimestampsForSurah` wraps `fetchAndCache` in try/catch, but `hasTimestampSource -> hasSource` (and `hasSurah`) call `localSurahs()` with no catch anywhere in the chain (`TimestampService.ts:15-17`). A fork whose `timestampLocalSurahList` throws would propagate uncaught into the follow-along UI. This can't affect Bayaan (field unset), but guarding fork-supplied code is the whole point of these seams, and the asymmetry with the already-guarded data path is worth closing:

```ts
private localSurahs(rewayatId: string): number[] {
  try { return branding.timestampLocalSurahList?.(rewayatId) ?? []; }
  catch { return []; }
}
```

Nits: `localSurahs(rewayatId)` is evaluated twice on a fall-through `fetchAndCache` (local block + `hasSurah`): fine since the provider is contracted side-effect-free, but a local `const` would avoid the second call. And worth a one-line note to fork authors that a `'local'`-sourced surah is immutable once cached (`isSurahCached` ignores `source`), so shipping an updated bundle needs a cache reset.

omar-zarka:
Thanks Osman — all three addressed in `a4b2993`:

- **Should-fix (unguarded coverage path).** `localSurahs()` now wraps `timestampLocalSurahList?.(rewayatId) ?? []` in try/catch → `[]` on throw, matching the already-guarded `timestampLocalProvider` data path. A throwing fork provider can no longer propagate into the follow-along UI, and the asymmetry you flagged is closed.
- **Nit (double provider call on fall-through).** Hoisted `const localList = this.localSurahs(rewayatId)` in `fetchAndCache` and split the R2 logic into a private `r2HasSurah(rw, surah)`. The R2 guard is now `!localList.includes(surah) && !this.r2HasSurah(this.findRewayat(...), surah)` — logically identical to `!hasSurah(...)` (`!(A || B)`), but `timestampLocalSurahList` is invoked exactly once per `fetchAndCache` instead of twice. `hasSurah` keeps the same public behavior (local check + `r2HasSurah`).
- **Nit (immutable once cached).** Added a `NOTE (fork authors)` at the `'local'` write: `isSurahCached` keys on (rewayat, surah) and ignores `source`, so a subsequently shipped/updated bundle needs a timestamp-cache reset to take effect.

Behavior-preserving; still byte-equivalent when the seam is unset.

PR #306, opened 2026-06-19
feat(rfc-018): community-reflections provider + render-slot seam

omar-zarka:
Implements the code for **RFC-018** (#299) — shape approved ("Approving the shape with these comments"). Adds two paired, optional, nullable `Branding` fields so a fork can surface community-authored ayah reflections **without forking the hot verse-render path**. **Both unset is Bayaan's default → the surface stays off and the Mushaf renders byte-identically to develop (zero behavior change).** Fifth in the XF-NNN series (after RFC-007/008/009/013).

## The seam
- `types/CommunityReflection.ts` (new) — `CommunityReflection` + `CommunityReflectionsProvider` type.
- `config/branding.d.ts` — `communityReflectionsProvider?` (data) + `ayahCommunityReflectionsComponent?: ComponentType<{surahNumber, ayahNumber}>` (render slot, RFC-008 pattern). Both `undefined` for Bayaan; `config/branding.js` documents the seam (0 behavior diff).

## Wired (all gated; no-op when unset)
- `components/mushaf/AyahCommunityReflections.tsx` (new) — default-noop wrapper mounted under the Arabic line in `VerseItem`; returns `null` when no slot is wired, otherwise renders the fork's component behind a **silent per-ayah error boundary** (resets on verseKey change for FlashList recycling).
- `store/mushafSettingsStore.ts` — `showCommunityReflections` opt-in (default false) + persist v16→v17 (additive, default false).
- `components/MushafSettingsContent.tsx` — toggle row rendered only when a provider is wired (no orphan toggle for Bayaan).
- `components/sheets/VerseActionsSheet.tsx` + `verse-actions/CommunityReflectionsContent.tsx` (new) — gated EXPLORE row + a generic fullscreen popup driven by the same provider field.

## Review tightening (from #299)
1. **Coupling / double-fetch:** the wrapper does not fetch; the fork's component owns the provider call + caching → single fetch site.
2. **Blast radius:** `VerseItem` renders in `QuranView`, `ContinuousListView`, and `ReadingPageView`; the unconditional mount appears in all three (acknowledged in a mount-site comment), kept free by the noop + the toggle/provider gates.
3. **Hot path:** non-blocking; a silent ErrorBoundary at the mount site so a provider/render throw renders nothing rather than crashing the recycling row.

## Verification
`npx tsc --noEmit` clean. The two modified settings files are kept to the RFC-018 surface only (no drive-by reformatting of develop's pre-existing prettier debt).


osmansaeday (changes requested):
All three findings from the doc review are addressed in code: the wrapper doesn't fetch (the fork component owns the provider call -> single fetch site), the mount-site comment in `VerseItem` acknowledges all three callers, and `SilentReflectionsBoundary` is a correct class boundary (`getDerivedStateFromError` + `componentDidUpdate` resetting on `verseKey` change for FlashList recycling). The v16->v17 persist migration is additive and correct, and the loading/error/empty triage in the popup is complete. Nice work on the boundary.

A few things before merge:

- **`theme.colors.primary` in new code** (`CommunityReflectionsContent.tsx`, verified-badge checkmark). This repo deprecates `theme.colors.primary`/accent colors in new code; the rest of this file already uses `theme.colors.text` with alpha, so match that (e.g. `Color(theme.colors.text).alpha(0.6)`).
- **Hot-path hook runs before the static gate** (`AyahCommunityReflections.tsx`). `useMushafSettingsStore(s => s.showCommunityReflections)` is called unconditionally, then `if (!SlotComponent) return null` runs after. `branding.ayahCommunityReflectionsComponent` is a module constant (always `undefined` on Bayaan), so every Bayaan user pays a Zustand subscription per verse row on the recycling FlashList; the PR's "free no-op when unwired" claim isn't quite true. Rules-of-hooks means you can't reorder in place: split into an outer component that checks the static constant first and returns `null`, delegating to an inner component that runs the hook only in fork builds.
- **`onBack` is a dead prop.** `CommunityReflectionsContentProps` declares `onBack: () => void` and the call site passes it, but the component destructures only `{surahNumber, ayahNumber}`. The parent sheet's back row handles navigation, so either drop `onBack` from the interface + call site, or wire it.
- **Toggle label overpromises.** The settings row gates on `communityReflectionsProvider` alone, but the inline slot needs the component too; a fork shipping only a provider gets a toggle whose "under each ayah" description does nothing (the action-sheet row still works). Either gate the toggle on both fields or make the label provider-scoped.

Nits: `ReflectionRow` is a `Pressable` even when `url` is empty (add `disabled={!url}`); no `accessibilityRole="link"`/label on the row (consistent gap with existing sheets, but worth adding here).

omar-zarka:
Thanks Osman — all four + the nits addressed in `48aef0d`:

- **Hot-path hook before the static gate.** `AyahCommunityReflections` now runs only the module-constant checks first (`if (!branding.ayahCommunityReflectionsComponent) return null; if (!branding.communityReflectionsProvider) return null;`) and delegates to a fork-only inner `AyahCommunityReflectionsSlot` that runs the `useMushafSettingsStore(s => s.showCommunityReflections)` subscription. Stock Bayaan (neither wired) returns before any hook, so a recycling FlashList row pays no per-row Zustand subscription; only forks that ship the slot pay it. Same overall gate (component AND provider AND toggle), reordered across the component boundary so rules-of-hooks holds.
- **`theme.colors.primary` in new code.** The verified-badge checkmark now uses `Color(theme.colors.text).alpha(0.6).toString()`, matching the rest of the file.
- **`onBack` dead prop.** Dropped from `CommunityReflectionsContentProps` and from the call site in `VerseActionsSheet` — the parent sheet's back row owns navigation.
- **Toggle overpromises.** The settings row now gates on BOTH `communityReflectionsProvider` and `ayahCommunityReflectionsComponent`; its "under each ayah in list view" copy describes the inline slot specifically, so a provider-only fork no longer gets a toggle whose description does nothing (the action-sheet surface is unaffected). Comment updated to spell out the two-field gate.
- **Nits.** `ReflectionRow`'s `Pressable` is now `disabled={!reflection.url}` with `accessibilityRole="link"` + an `accessibilityLabel`.

tsc + prettier (3.8.1) clean; eslint 0 errors on the touched files.

osmansaeday:
Thanks for turning this around, the earlier round is fully addressed and it's close. Three small things before I re-approve, then this is good to go.

**1. Security: validate the URL scheme before opening it.** `ReflectionRow.handleOpen` (`components/sheets/verse-actions/CommunityReflectionsContent.tsx:458`) passes third-party, community-authored `reflection.url` straight to `Linking.openURL` with no scheme check. Because a fork's provider is the only thing between arbitrary content and this call, an `intent://` (Android) or other non-web scheme could be smuggled through. Please gate it, e.g.:

    if (/^https?:\/\//i.test(reflection.url)) Linking.openURL(reflection.url);

This is a dead path for Bayaan today, but it ships to every fork that wires the seam, so worth hardening now.

**2. Generalize the hardcoded provider name.** The accessibility label at `:465` bakes in "QuranReflect", but this file's own header frames it as a generic, provider-agnostic surface. Any non-QuranReflect provider would read wrong. Suggest `` `Open ${reflection.author.name}'s reflection` ``.

**3. `locale` is never passed to the provider** in the popup path (`:459`), even though `CommunityReflectionsProvider`'s third param exists for the user's translation preference. Either thread the current locale through or drop a one-line note on why it's intentionally omitted here.

Sequencing: the RFC-018 doc (#299) has now landed on `develop`, so the JSDoc references resolve. Once the three above are in, ping me for a fresh approval (GitHub still shows my stale CHANGES_REQUESTED from before 48aef0d).

omar-zarka:
Thanks Osman — merged `develop` in (`b1effb7`, one trivial import conflict in `branding.d.ts`; RFC-018 + RFC-019 fields coexist, and the v17 persist migration still lands cleanly on top of develop's v16), and all three follow-ups are in `190ba9d`:

1. **URL scheme validation.** `ReflectionRow` now gates on `/^https?:\/\//i` before `Linking.openURL` — an `intent://`, `javascript:`, or `file:` smuggled through a provider's community-authored `url` never reaches the opener. The row's `Pressable` is also `disabled` for non-web URLs, so a bad URL degrades to an inert card rather than a dead tap.
2. **Provider-agnostic label.** The accessibility label is now `` `Open ${reflection.author.name}'s reflection` `` — no baked-in "QuranReflect".
3. **`locale`.** Went with the note: stock Bayaan has no reliable BCP-47 source to thread — `selectedTranslationId` is a translation-*edition* id (the bundled editions carry `language: 'English'` prose, and the remote edition registry is itself fork-overridable via RFC-009), so deriving a tag from it would be wrong as often as right. The call site now documents that we lean on the contract's `undefined → provider's own default` (the reference provider defaults to `'en'` and owns its language filtering). If a first-class user-language preference ever lands, threading it here is a one-liner.

tsc + prettier (3.8.1) + eslint clean; full jest suite green on the merged tree. Ready for a fresh look whenever you are.

osmansaeday (approved):
Re-reviewed at head `b889507d9`. All four items from my Jun 23 review are addressed:

- Deprecated `theme.colors.primary` on the verified badge replaced with `Color(theme.colors.text).alpha(0.6)` (no `primary`/accent left in the PR).
- Hot-path hook split into an outer component that checks the static branding constants first (no hooks, returns `null` on Bayaan) delegating to an inner slot that owns the Zustand subscription, so Bayaan pays nothing per verse row.
- Dead `onBack` prop removed from the interface and the call site.
- Settings toggle now gated on both `communityReflectionsProvider` and `ayahCommunityReflectionsComponent`.

Nits were fixed and hardened beyond the ask (`disabled` on non-web URLs, `accessibilityRole="link"` + label, plus a `^https?://` guard on `Linking.openURL` that closes a scheme-injection vector from fork data). The v16 to v17 persist migration is additive and safe, gating has no orphan-toggle gaps, and zero-behavior-change for Bayaan is verified through all three `VerseItem` call sites.

One cosmetic nit remains (a one-frame blank after the error boundary resets on `verseKey` change) but it is not worth blocking. Approving.

PR #307, opened 2026-06-19
feat: add download-all for a reciter's surahs

bayzhan8:
## What does this PR do?

Adds a **download-all** button to the reciter profile action bar (next to Favorite / Play / Shuffle). One tap downloads every surah for the **currently selected rewayah** for offline playback, so listeners who follow a particular reciter no longer have to open each surah and download it individually.

The button has three states:
- Idle: download icon (only shows the percentage/checkmark states once acted on).
- In progress: shows aggregate percentage; tap again to cancel.
- All downloaded: checkmark.

It targets the same surah set as Play All / Shuffle All (current rewayah, respecting the loved/search filters), for consistency.

### Implementation notes
- New dependency-injected helper `services/player/bulkDownloadSurahs.ts`. It imports nothing from expo/React Native, which keeps the bulk-download logic unit-testable in isolation. It mirrors the existing bulk-download flow used elsewhere in the app: skips already-downloaded and in-flight items, throttles progress writes, spaces downloads out to avoid AsyncStorage write storms, and isolates per-item failures so one bad download doesn't abort the batch.
- Download-tracking IDs use the exact format `SurahItem` reads (`reciterId-surahId-rewayatId`), so the existing per-row progress rings and downloaded checkmarks animate live during the batch.
- `ActionButtons` gains optional props (`onDownloadAllPress`, `isDownloadingAll`, `downloadAllProgress`, `allDownloaded`), all with safe defaults so existing usage is unaffected.
- No changes to the existing single-surah, loved, or playlist download paths.

Closes #

## Type of change

- New feature

## Testing

- Added unit tests (`services/player/__tests__/bulkDownloadSurahs.test.ts`) covering: downloading the pending set, skipping already-downloaded items, skipping in-flight items, continuing past per-item failures, cancellation, and aggregate progress reporting. All 9 pass.
- `npx tsc --noEmit` passes; `npx prettier --check` and `eslint` are clean on the changed files.
- [ ] Tested on iOS
- [ ] Tested on Android

> Note: on-device iOS/Android testing has not been run yet. The new logic is covered by unit tests and the UI reuses the app's existing, shipped download plumbing. Happy to validate on simulators/devices or adjust the UX (icon/placement/copy) per maintainer preference.

## Checklist

- [x] Ran `npx prettier --write` on all changed files
- [x] Ran `npx tsc --noEmit` — no new type errors
- [x] Branch targets `develop` (not `main`)
- [x] No `console.log` debug statements left in
- [x] AI assistance used? If yes, all AI-generated blocks are marked with `// @ai`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/components/reciter-profile/ReciterProfile.tsx
+++ b/components/reciter-profile/ReciterProfile.tsx
@@ -41,7 +41,7 @@ import {useFavoriteReciters} from '@/hooks/useFavoriteReciters';
 import {
   useDownloadQueries,
   useDownloadActions,
-  useDownloads,
+  useAreAllSurahsDownloaded,
 } from '@/services/player/store/downloadSelectors';
 import {downloadSurah} from '@/services/downloadService';
 import {bulkDownloadSurahs} from '@/services/player/bulkDownloadSurahs';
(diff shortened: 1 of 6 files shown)


OKKHALIL3:
Pushed a follow-up commit (ea6aba39) directly to address review feedback on the bulk download-all flow, so this stays unblocked:

- **Narrowed the store subscription** — `allDownloaded` now comes from a new primitive-returning selector (`useAreAllSurahsDownloaded`) instead of subscribing to the whole `downloads` array, so the (heavy) reciter-profile screen only re-renders when the boolean flips, not on every single-surah download / removal / batch item.
- **Unmount guard** — an `isMountedRef` gates the post-await `setState`/`Alert` calls, and unmount now cancels the in-flight batch, so navigating away mid-batch no longer pops a completion alert on an unrelated screen or sets state after unmount.
- **Double-tap reentrancy** — an in-flight ref closes the same-frame window so two rapid taps can't start two concurrent batches.

CI (Jest + tsc) is green. Thanks for the feature! 🙏

PR #310, opened 2026-06-30
fix(settings): stop the closed BottomSheetModal from eating all touches on slow Android

omar-zarka:
### Problem

The shared `components/BottomSheetModal.tsx` wrapper (the confirm / note bottom sheets) can leave its **host screen completely touch-dead on slow Android CPUs** while the sheet is logically closed — taps register nowhere, even though the rest of the app keeps working. This is the same class as the PlayerSheet freeze fixed in #308, on a separate ungated component instance.

### Root cause

gorhom's `BottomSheetBackdrop` mounts with `pointerEvents:'auto'` (a full-screen, opacity-0-but-hit-testable Tap `GestureDetector`) and only flips to `'none'` via a reanimated `useAnimatedReaction` → `runOnJS` → `setState` chain gated on its internal `animatedIndex` settling to `-1`. On slow CPUs the closed-sheet layout race delays that settle, so the backdrop keeps eating every touch while the sheet is **logically** closed.

And — exactly as #308 found — the **elevated BottomSheet Body is a SEPARATE sibling** (`elevation/zIndex: 3000`) that drops its touch-interactivity asynchronously too. Gating only the backdrop is not enough; the Body still swallows touches.

### Fix

Gate **both** siblings on the logical `isVisible` prop (immune to gorhom's racing `animatedIndex`):

- `renderBackdrop` returns `null` when `!isVisible`
- the `BottomSheet` `style` adds `pointerEvents: 'none'` when `!isVisible`

When the sheet is open, behavior is identical to today.

### Why two-sibling

#308 (PlayerSheet) is the precedent: an earlier single-sibling attempt there still froze because the elevated Body was a second touch-eater. This mirrors that complete fix on the `BottomSheetModal` wrapper.

### Validation

This is a faithful port of a fork-side fix that's **tester-confirmed** on the original slow-CPU device (Redmi Note 9 / Helio G85 class) and passed `tsc` + unit tests + a release boot-gate. The diff is additive-only (no behavior change when open); `prettier` (3.8.1, lockfile-pinned) reports the file unchanged after formatting.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/components/BottomSheetModal.tsx
+++ b/components/BottomSheetModal.tsx
@@ -86,14 +86,29 @@ const BottomSheetModal: React.FC<BottomSheetModalProps> = ({
   }, [isVisible, onClose]);
 
   const renderBackdrop = useCallback(
-    (props: BottomSheetBackdropProps) => (
-      <BottomSheetBackdrop
-        {...props}
-        disappearsOnIndex={-1}
-        appearsOnIndex={0}
-      />
-    ),
-    [],
+    (props: BottomSheetBackdropProps) => {
+      // gorhom's BottomSheetBackdrop mounts with pointerEvents:'auto' (a
+      // full-screen, opacity-0-but-hit-testable Tap GestureDetector) and only
+      // flips to 'none' via a reanimated useAnimatedReaction -> runOnJS ->
+      // setState chain gated on its internal animatedIndex settling to -1. On
+      // slow Android CPUs the closed-sheet layout race delays that settle, so
+      // the backdrop keeps eating every touch while the sheet is LOGICALLY
+      // closed -> the entire host screen goes touch-dead until the next
+      // re-render. Gate on the logical `isVisible` prop so the backdrop never
+      // renders while closed, immune to gorhom's racing animatedIndex. Same
+      // touch-eater fix shipped for the PlayerSheet backdrop in #308.
+      if (!isVisible) {
+        return null;
+      }
+      return (
+        <BottomSheetBackdrop
+          {...props}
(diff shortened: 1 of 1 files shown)


osmansaeday:
The fix itself is a clean, faithful mirror of #308 (which has now merged), so no concerns on the mechanics. One thing to confirm before this lands:

`components/BottomSheetModal.tsx` has no import sites anywhere in Bayaan (I checked the working tree, all worktrees, and develop). So this is prophylactic for the downstream fork rather than fixing an active in-repo freeze, which is fine, I just want to confirm that's the intent and the file is deliberately kept rather than a removal candidate.

Also, same cosmetic footnote as #308: closing via a button/back (the norm for a confirm sheet) now drops the backdrop scrim instantly instead of fading, since the unmount is tied to the `isVisible` flip that precedes the close animation. Purely visual, doesn't reintroduce the touch-eating. Worth a quick look on the fork where it actually renders.

Assuming it's intentional downstream code, happy to merge.

osmansaeday (approved):
Approving. Faithful mirror of the #308 PlayerSheet fix, now covering both racing siblings: the backdrop (returns `null` while `!isVisible` instead of waiting on gorhom's `animatedIndex → runOnJS → setState` settle) and the elevated body (`pointerEvents` gated on `isVisible`). Gating both on the logical `isVisible` prop rather than gorhom's async `animatedIndex` is the right way to kill the closed-sheet touch-eater on slow Android. Clean and minimal.

PR #294, opened 2026-05-31
feat: upgrade Expo SDK 55 → 56

OKKHALIL3:
## Summary

- Upgrades all Expo packages from SDK 55 to SDK 56 (including `expo-router`, `expo-audio`, `expo-image`, and all related packages)
- Migrates `@react-navigation` direct imports to their `expo-router` re-exports (SDK 56 ships these under its own namespace)
- Removes deprecated `StatusBar` props (`translucent`, `backgroundColor`) and replaces `StyleSheet.absoluteFillObject` with `StyleSheet.absoluteFill`
- Updates all Android and iOS native project files (Gradle, Podfile, Xcode, manifests, entitlements)
- Enables React Compiler in `app.config.js`
- Removes unused Play Store metadata, sentry config files, and `ExportOptions.example.plist`
- Inlines `EventSubscription` type in audio services to avoid importing from `expo-modules-core`
- Adds `upload-adhkar-audio` script

## Commits (21)

1. `chore(deps)`: upgrade Expo SDK 55 → 56 and bump peer packages
2. `feat(config)`: enable React Compiler in app.config.js
3. `fix(imports)`: replace `@react-navigation/native` with expo-router re-exports
4. `fix(imports)`: replace `@react-navigation/elements` with `expo-router/react-navigation`
5. `fix(imports)`: replace `@react-navigation/bottom-tabs` with expo-router re-exports, drop deprecated StatusBar props
6. `fix(reciter-profile)`: update @react-navigation imports and remove deprecated StatusBar props
7. `fix(styles)`: replace `StyleSheet.absoluteFillObject` with `StyleSheet.absoluteFill`
8. `fix(audio)`: inline `EventSubscription` type instead of importing from expo-modules-core
9. `feat(scripts)`: add upload-adhkar-audio script
10. `chore(android)`: remove Google Play store metadata assets
11. `chore(android)`: upgrade Gradle wrapper for SDK 56
12. `chore(android)`: update build.gradle and gradle.properties for SDK 56
13. `chore(android)`: update AndroidManifest, styles, and gitignore for SDK 56
14. `chore(android)`: update splash screen assets for SDK 56
15. `chore`: remove sentry.properties from Android and iOS
16. `chore(ios)`: remove ExportOptions.example.plist
17. `chore(ios)`: update Podfile and Podfile.properties for SDK 56
18. `chore(ios)`: update Podfile.lock for SDK 56
19. `chore(ios)`: update Xcode project for SDK 56
[...]

PR #311, opened 2026-07-04
chore: pin @gorhom/bottom-sheet ^5.2.14 (pre-empt SDK-56 / reanimated-4.3 regression)

omar-zarka:
## What

Raise the `@gorhom/bottom-sheet` floor from `^5.2.8` → `^5.2.14` (plus the matching single lockfile entry).

## Why

On the **Expo SDK 56 stack** (`react-native-reanimated` 4.3.1 / RN 0.85, new architecture — i.e. **`develop` since #294**), gorhom **5.2.8**'s imperative `BottomSheet.expand()` / `snapToIndex()` become **silent no-ops** — the ref is valid and exposes the methods, the calls run with **no error**, but the sheet's open animation never fires. Not z-order, not the snap-point type, not `FullWindowOverlay`; the gorhom open animation itself is dead on 5.2.8 under reanimated 4.3. **≥5.2.14 fixes it.**

I hit this in a downstream SDK 56 migration: it broke the mini-player → full-player expand (the tap *was* firing — play/pause on the same control worked — the sheet just never presented) and cost a build plus a few hours to bisect. The bisection that isolated it: force-call `expand()` from a temp effect over Metro and flip only the gorhom version.

## Scope / risk

- **Correction to my original framing:** `develop` has been on SDK 56 / reanimated 4.3.1 since #294 merged (2026-06-29), so this is an **active fix, not forward-looking** — any imperative `expand()`/`snapToIndex()` on `develop` today silently dies on 5.2.8. (I first opened this reading `main`, which is still SDK 55, and wrongly assumed `develop` matched.)
- Pure JS, **within the existing `^5.2.8` caret** — no native rebuild.
- 5.2.14 has **identical `dependencies` + `peerDependencies`** as 5.2.8 (`@gorhom/portal` 1.0.14, `invariant` ^2.2.4), so the lockfile delta is just this one package's `version`/`resolved`/`integrity`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@expo/server": "^0.1.0",
         "@expo/vector-icons": "^15.0.3",
         "@fawazahmed/react-native-read-more": "^3.0.4",
-        "@gorhom/bottom-sheet": "^5.2.8",
+        "@gorhom/bottom-sheet": "^5.2.14",
         "@legendapp/list": "^1.0.5",
         "@miblanchard/react-native-slider": "^2.6.0",
         "@react-native-async-storage/async-storage": "2.2.0",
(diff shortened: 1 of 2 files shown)


omar-zarka:
Correction: I originally framed this as forward-looking, but `develop` has been on SDK 56 / reanimated 4.3.1 since #294 (2026-06-29) — so on `develop` today gorhom 5.2.8's `expand()`/`snapToIndex()` silently no-op. This is an **active** fix, not a pre-emptive one. Updated the description accordingly.

PR #312, opened 2026-07-06
fix(mushaf): preserve page across rotation via orientation key-remount

omar-zarka:
## What

An orientation `key` on the Mushaf reader list(s) so they remount on a portrait↔landscape flip. One-line-per-list change (+12 total: the `orientationKey` const + a `key` on each of the four reader branches).

## Why

The default Mushaf reader is a paged, **`inverted`** horizontal `FlatList` whose `getItemLayout` sizes each item at `metrics.screenWidth`. On rotation the item width changes, but the native list keeps its **pixel** `contentOffset` → the visible page drifts to a different index (old offset ÷ new item width). `initialScrollIndex` is **mount-only** and never re-fires, so nothing pulls it back. **Rotating portrait↔landscape throws the reader onto a different surah** (a tester lost their place and landed ~23 pages away).

## Fix

A `key={windowWidth > windowHeight ? 'landscape' : 'portrait'}` on all four reader branches (both continuous views + both horizontal FlatLists). On an orientation flip the list **remounts fresh**, starting at `initialScrollIndex = currentPage`, so it never inherits the stale offset — the drift can't happen. Keyed on **orientation**, not raw dimensions, so the on-screen keyboard shrinking the window height never triggers a needless remount.

### Why not `maintainVisibleContentPosition`?

Tried it first — it's the "natural" native fix. It does **not** anchor an inverted, paged list across a rotation *resize*: on-device the content drifted to a different surah while the header stayed correct. A JS `scrollToIndex` re-anchor is a frame late (flashes the wrong page), and masking that flash only treats the symptom. The key-remount prevents the drift at the source.

## Verification

Verified on a **Pixel 3 (release build)** in a downstream fork whose Mushaf reader is byte-identical to this file apart from this change (same SDK 56 / reanimated 4.3 stack as `develop`):
- Pages held across **portrait → landscape → portrait** (tested on pages 50 and 49).
- Transition settles within **~300 ms with no wrong-page frame**.
- Normal page turns unaffected (50→49 swipe).
- `tsc` clean · jest green · prettier clean.

Pure JS, no native change; applies identically on SDK 55 (`main`) since the drift is an RN offset-preservation behavior, not stack-specific.

### DIFF
--- a/components/mushaf/main.tsx
+++ b/components/mushaf/main.tsx
@@ -541,6 +541,14 @@ export default function MushafViewer({
   const insets = useSafeAreaInsets();
   const {isTablet} = useResponsive();
   const {width: windowWidth, height: windowHeight} = useWindowDimensions();
+  // A paged, `inverted` FlatList preserves its pixel scroll offset when its
+  // items resize on rotation, so the visible page drifts to a different index
+  // (old offset ÷ new item width); `initialScrollIndex` is mount-only and never
+  // re-fires. Remount the list on an orientation flip via this key so it starts
+  // fresh at `initialScrollIndex = currentPage` and never inherits the stale
+  // offset. Keyed on orientation (not raw dimensions) so the on-screen keyboard
+  // shrinking the height never remounts the reader.
+  const orientationKey = windowWidth > windowHeight ? 'landscape' : 'portrait';
 
   // Measured size of the FlatList container. Using `useWindowDimensions()`
   // alone overestimates the available space on iPad because the stack
@@ -1036,6 +1044,7 @@ export default function MushafViewer({
       {/* Content area: horizontal FlatList or vertical continuous view */}
       {isVertical && viewMode === 'mushaf' ? (
         <ContinuousMushafView
+          key={orientationKey}
           ref={continuousListRef}
           textColor={readingColors.text}
           dividerColor={readingColors.textSecondary}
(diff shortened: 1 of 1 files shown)


osmansaeday:
Nice fix, re-deriving `initialScrollIndex` from `currentPage` on a key-remount is the right call and I verified it lands on the correct page. One real issue and one nit.

**Analytics double-count on rotation.** The horizontal `FlatList`'s `onViewableItemsChanged` (`components/mushaf/main.tsx:814-835`) has no equality guard, so the forced remount on every rotation re-fires it for the same page once it settles, bumping `pagesOpenedRef` and restarting the 15s `page_read` timer even though the user never left the page. This will inflate `mushaf_page_opened` / pages-read counts. The vertical path already guards this (`if (page !== currentPage)` in `handleContinuousPageChange`), so please mirror that here.

**Nit: orientation predicate mismatch.** `orientationKey` uses `windowWidth > windowHeight` while `getMushafLayout` derives `facingPages` from `width >= height` (`constants.ts:162`). At an exact-square viewport they can disagree and skip the remount this PR relies on. Very low real-world odds, but aligning both to `>` (or both `>=`) removes the edge case.

Happy to merge once the analytics guard is in; the predicate is optional.

osmansaeday (approved):
Approving. Root cause is nailed: a paged/inverted FlatList keeps its pixel scroll offset when items resize on rotation, so the visible index drifts and `initialScrollIndex` (mount-only) never re-fires. Remounting via `key={orientationKey}` forces a fresh mount at `initialScrollIndex = currentPage`, which is the correct fix. Good call keying on orientation (portrait/landscape) rather than raw dimensions, so the on-screen keyboard shrinking the height can't trigger a spurious remount. Applied consistently across all four list variants. CI green.

PR #313, opened 2026-07-12
chore(types): environment-agnostic timer-handle types for RN 0.85

omar-zarka:
RN 0.85 (SDK 56) types `setTimeout`/`setInterval` returns as `number` on Hermes. Five timer-handle sites were still typed `NodeJS.Timeout` — they only typecheck today because `@types/node`'s globals win the resolution under tsconfig `"types": ["node"]`, and that type lies about the runtime handle (e.g. `.refresh()`/`.unref()` typecheck but would crash on Hermes's numeric handle). This switches them to `ReturnType<typeof setTimeout>` / `ReturnType<typeof setInterval>`, which is correct under any resolution order and matches the pattern the SDK 56 tree already uses elsewhere (`AmbientAudioService`, `MushafAudioService`, `PerformanceMonitor`, `mushaf/main`).

Type-level only — no behavior change, no formatting drift (5 files, 5 lines). Mirrors the fix Qariah shipped during its SDK 56 absorption, where these exact sites surfaced as type breaks. `tsc --noEmit` and the full jest suite are green.

### DIFF
--- a/components/player/v2/PlayerSheet.tsx
+++ b/components/player/v2/PlayerSheet.tsx
@@ -94,7 +94,7 @@ export const PlayerSheet = () => {
 
   // Effect to handle sleep timer remaining time
   useEffect(() => {
-    let interval: NodeJS.Timeout;
+    let interval: ReturnType<typeof setInterval>;
     if (settings.sleepTimerEnd) {
       interval = setInterval(() => {
         // Calculate remaining time based on the end timestamp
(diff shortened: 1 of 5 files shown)


osmansaeday (approved):
Approving. Type-only fix: `NodeJS.Timeout` → `ReturnType<typeof setTimeout>`/`ReturnType<typeof setInterval>` at all 5 timer-handle sites. `NodeJS.Timeout` only typechecks because `@types/node` globals win resolution, and it misrepresents the Hermes handle (a number) so `.refresh()`/`.unref()` would typecheck then crash. `ReturnType<typeof …>` is correct under any resolution order and matches the pattern already used elsewhere in the SDK 56 tree. Each site correctly matches setInterval vs setTimeout. CLEAN, CI green, no behavior change.

PR #314, opened 2026-07-18
fix(mushaf): paint bookmarked verses persistently + load annotations in the paged pipeline

omar-zarka:
## What

Bookmarking an ayah in the mushaf currently leaves **no visible trace on the page** — the momentary highlight users see is the *selection* layer clearing when the action sheet closes. Users read that as "my highlight didn't save" (we got exactly this report on the Qariah fork; the mechanics are identical on develop). This PR:

1. **Paints bookmarked verses persistently** in both Skia pipelines (`SkiaPage` + `ContinuousMushafView`) via a new "Layer 0.5" using `BOOKMARK_HIGHLIGHT_COLOR` (the existing yellow highlight token). Priority is deliberate: explicit colored highlights, playback, and selection all paint **over** it, and the themes layer skips bookmarked verses — so nothing existing changes appearance except bookmarks gaining a trace.
2. **Fixes the paged pipeline never loading the annotations store.** Only the continuous views and the player QuranView call `loadAnnotationsForSurah`; the horizontal paged pipeline never does. Consequences on develop today, independent of the paint: on cold entry the long-press sheet reports **stale bookmark state** ("Bookmark" on an already-bookmarked verse), and tapping it **INSERTs a duplicate row** (`VerseAnnotationDatabaseService.addBookmark` has no dedupe). A new page-level effect in `main.tsx` loads every surah on the current page.
3. **Store upgrade — `loadAnnotationsForSurahs(number[])`:**
   - **Multi-surah aware.** A single mushaf page can span several surahs (the norm in Juz 'Amma); the single-surah cache blinds the renderer to bookmarks in the page's later surahs.
   - **No dropped loads.** The old `if (loading) return` guard silently discarded a concurrent caller's surah set; loads now serialize through an in-flight promise and re-check the cache key after waiting.
   - **Subset no-op.** `loadAnnotationsForSurah(n)` no-ops when `n` is already inside the loaded set, so a view-level per-surah load can't narrow a page-level superset mid-scroll.
4. **6 regression tests** (`store/__tests__/verseAnnotationsStore.test.ts`): cross-surah merge, exact-key + subset no-ops, in-flight serialization, failure path leaves the previous key intact, optimistic mutations.

## Verification

- `tsc --noEmit` clean · **jest 142/142** (6 new) · prettier **3.8.1** (lockfile-pinned) clean on all touched files.
- **Runtime device-verified on a Pixel 3 (release build)** via the Qariah fork, whose `SkiaPage.tsx`, `ContinuousMushafView.tsx`, and `verseAnnotationsStore.ts` were **byte-identical to develop** before this change (faithful-proxy caveat: `main.tsx` carries small fork divergences, so its 24-line effect was re-applied here on develop's copy and is covered by tsc/tests, not the device run). Verified on device: cold-open paint straight from the DB, bookmark → tint persists after the sheet closes, remove → clears live, **multi-surah page 604** (bookmark on An-Nas, the 3rd surah on the page, paints), cold-relaunch persistence, cross-page repaint, vertical/continuous mode, segment-exact extent (the tint stops at the verse marker; the next ayah is untinted).

## Notes for review

- With `coloredHighlights` enabled (Bayaan default), an explicit colored highlight on a bookmarked verse **wins** — the bookmark tint only shows where the user hasn't highlighted.
- The paint is intentionally **not feature-flag-gated**: no in-component flag consumption exists in the codebase today, and the change is a fix for "bookmarks leave no trace." If you'd rather ship it dark or behind a flag, happy to adjust.
- `addBookmark`'s missing dedupe at the DB layer is *masked* by the loader fix (the sheet now shows correct state, so users can't re-add). An upsert there would be cheap defense-in-depth — left out to keep this focused; can add on request.

### DIFF
--- a/components/mushaf/main.tsx
+++ b/components/mushaf/main.tsx
@@ -49,6 +49,8 @@ import {mushafSessionStore} from '@/services/mushaf/MushafSessionStore';
 import {useMushafNavigationStore} from '@/store/mushafNavigationStore';
 import {useMushafVerseSelectionStore} from '@/store/mushafVerseSelectionStore';
 import {useMushafPlayerStore} from '@/store/mushafPlayerStore';
+import {useVerseAnnotationsStore} from '@/store/verseAnnotationsStore';
+import {mushafVerseMapService} from '@/services/mushaf/MushafVerseMapService';
 import {useMushafAutoPageTurn} from '@/hooks/useMushafAutoPageTurn';
 import {MushafPlayerBar} from './MushafPlayerBar';
 import SkiaPage from './skia/SkiaPage';
@@ -636,6 +638,28 @@ export default function MushafViewer({
     : {};
 
   const currentSurahId = pageToSurah[currentPage] || 1;
+
+  // @ai — load verse annotations for every surah on the current page.
+  // Nothing in the horizontal paged pipeline loaded the store before, so
+  // bookmark tints never painted and the long-press sheet showed stale
+  // bookmark state on cold entry (tapping it then INSERTed a duplicate row).
+  // Multi-surah aware because Juz-'Amma pages hold several surahs; the
+  // continuous views' own per-surah loads become subset no-ops against this.
+  const loadAnnotationsForSurahs = useVerseAnnotationsStore(
+    s => s.loadAnnotationsForSurahs,
+  );
+  useEffect(() => {
+    const verseKeys =
+      mushafVerseMapService.getOrderedVerseKeysForPage(currentPage);
+    let surahs = [
+      ...new Set(verseKeys.map(vk => Number(vk.split(':')[0]))),
+    ].filter(n => Number.isFinite(n) && n > 0);
+    if (surahs.length === 0 && digitalKhattDataService.initialized) {
+      const fallback = digitalKhattDataService.getPageToSurah()[currentPage];
+      if (fallback) surahs = [fallback];
+    }
+    if (surahs.length > 0) void loadAnnotationsForSurahs(surahs);
+  }, [currentPage, loadAnnotationsForSurahs]);
 
   // --- Analytics: mushaf page tracking ---
   const pageReadTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
--- a/components/mushaf/skia/ContinuousMushafView.tsx
+++ b/components/mushaf/skia/ContinuousMushafView.tsx
@@ -29,7 +29,10 @@ import {useTajweedStore} from '@/store/tajweedStore';
 import {useMushafVerseSelectionStore} from '@/store/mushafVerseSelectionStore';
 import {useMushafPlayerStore} from '@/store/mushafPlayerStore';
 import {useVerseAnnotationsStore} from '@/store/verseAnnotationsStore';
-import {HIGHLIGHT_COLORS} from '@/types/verse-annotations';
+import {
+  BOOKMARK_HIGHLIGHT_COLOR,
+  HIGHLIGHT_COLORS,
+} from '@/types/verse-annotations';
 import {useTheme} from '@/hooks/useTheme';
 import {mushafPreloadService} from '@/services/mushaf/MushafPreloadService';
 import {
@@ -343,6 +346,11 @@ const MushafPageContent: React.FC<MushafPageContentProps> = React.memo(
 
     // ── Playback + annotation highlights ───────────────────
     const persistentHighlights = useVerseAnnotationsStore(s => s.highlights);
+    // @ai — bookmarked verses paint a persistent tint (same layer
+    // ordering as SkiaPage's paged pipeline).
+    const bookmarkedVerseKeys = useVerseAnnotationsStore(
+      s => s.bookmarkedVerseKeys,
+    );
     const playbackVerseKey = useMushafPlayerStore(s => {
       if (!s.currentVerseKey || s.playbackState === 'idle') return null;
       return s.currentVerseKey;
@@ -555,6 +563,7 @@ const MushafPageContent: React.FC<MushafPageContentProps> = React.memo(
       Map<number, Array<{start: number; end: number; color: string}>>
     >(() => {
       const hasAnnotations = Object.keys(persistentHighlights).length > 0;
+      const hasBookmarks = bookmarkedVerseKeys.size > 0;
       const hasPlayback = !!playbackVerseKey;
       const hasRewayahDiffs =
         showRewayahDiffs && rewayah !== 'hafs' && rewayahDiffService.hasDiffs;
@@ -563,7 +572,13 @@ const MushafPageContent: React.FC<MushafPageContentProps> = React.memo(
           ? new Set(selectedVerseKeys)
           : null;
 
-      if (!hasAnnotations && !hasPlayback && !selectedSet && !hasRewayahDiffs)
+      if (
+        !hasAnnotations &&
+        !hasBookmarks &&
+        !hasPlayback &&
+        !selectedSet &&
+        !hasRewayahDiffs
+      )
         return EMPTY_BG_MAP;
 
       const map = new Map<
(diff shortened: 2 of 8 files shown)


osmansaeday (changes requested):
Thanks for this. The layer design is the right approach (skip-based exclusivity, so nothing double-paints and there are no alpha-blend artifacts), it reuses the existing `getVerseSegmentsForPage` pipeline so the old Allah-name off-by-one doesn't recur, and there's no "re-render all pages" regression. A few things to fix before merge, one of them a real correctness bug.

**1. (blocker) Subset/superset cache race drops sibling-surah bookmarks on multi-surah pages.**
`loadAnnotationsForSurah(n)` reads `loadedKey` synchronously for its subset guard, so when the player's per-surah load fires while `main.tsx`'s page-level `loadAnnotationsForSurahs([112,113,114])` is still in flight, it doesn't see `n` as covered and delegates on. The post-wait recheck in `loadAnnotationsForSurahs` is exact-string (`get().loadedKey === key`), so `"112,113,114" === "113"` is false and it re-fetches surah 113 alone, then `set()` **replaces** the store, wiping 112/114 bookmarks and tints until a reload. Reachable via the common "read while listening" path on page 604.

**2. (should-fix) Same root cause causes tint flicker on surah-boundary pages, no concurrency needed.**
`windowSize={7}` keeps neighbor pages mounted, `setCurrentPage` fires at 50% visibility mid-swipe, and the effect reloads only the current page's surahs and replaces the store, so a still-mounted neighbor page spanning a different surah loses its tint. Most visible in Juz 'Amma, exactly where the feature should shine.

Findings 1 and 2 are fixed by one change: have the cache **merge/accumulate** loaded surahs (and make the recheck subset-aware) instead of replace + exact-match.

**3. (should-fix) `addBookmark` throws unhandled, and the description misstates the failure.**
`bookmarks.verse_key` is `TEXT NOT NULL UNIQUE` and `addBookmark` is a plain `INSERT` (no `OR IGNORE`), so a duplicate **throws** (SQLite ABORT), it does not insert a duplicate row. `handleToggleBookmark` has no try/catch, so any stale `isBookmarked` (including via finding 1) means the promise rejects and `store.addBookmark` / `SheetManager.hide` never run, leaving the sheet frozen. Suggest `INSERT OR IGNORE` plus a try/catch at the call site.

**4. (product) Bookmark tint is pixel-identical to the yellow highlight.**
`BOOKMARK_HIGHLIGHT_COLOR = HIGHLIGHT_COLORS.yellow`, so a bookmarked-only verse and an explicitly yellow-highlighted verse are indistinguishable. For a "give bookmarks a trace" feature that muddies the signal. Worth a distinct token, or confirm it's intended.

Nits: duplicate `pageNumber` in the `SkiaPage` `useMemo` deps; `loadedSurah` is now vestigial and misleading (no consumers, set to the smallest surah in the key); and the subset-no-op test awaits the superset load before issuing the subset call, so it doesn't actually cover finding 1. Reuse the `deferred()` gate from the serialization test to close that gap.

Really nice work on the paint priority and the segment reuse. The one that matters is finding 1, and the merge-based cache resolves 1 and 2 together.

omar-zarka:
Thanks — this was a genuinely useful review. All four findings reproduce; pushed in `8666c29`.

**1 + 2 (blocker + flicker) — fixed together, as you suggested, by making the cache accumulate.**

`loadedSurah` + `loadedKey` are replaced by `loadedSurahs: Set<number>` which accumulates and is never narrowed:

- `loadAnnotationsForSurah(n)` no-ops on `loadedSurahs.has(n)`.
- `loadAnnotationsForSurahs` no-ops when every wanted surah is already loaded; after `await inFlightLoad` it **recomputes** what is still missing rather than comparing an exact key, and returns when nothing is; only the missing surahs are fetched.
- Results **merge** into the store read at set-time (`get()` inside the setter, not a snapshot captured before the await), so a narrow load can never unpaint a wider one, and a neighbour page still mounted by `windowSize` keeps its tint across a `setCurrentPage`.

You were right that the exact-string recheck was the load-bearing part of the bug — the subset guard in `loadAnnotationsForSurah` reads the key synchronously, so it can't see an in-flight superset, and everything downstream followed from that.

**I confirmed the new test actually reproduces the old behaviour rather than assuming it.** Running the identical race against the pre-fix store ends with `loadedKey: '113'` and *both* sibling bookmarks gone (`has112: false, has114: false`). Which also confirms your point about the original subset test: because it awaited the superset before issuing the subset call, it could never have caught this. The new one holds the page-level load open with the existing `deferred()` gate and issues `loadAnnotationsForSurah(113)` while it is genuinely in flight.

**3 — you're right on the mechanism, and my description was wrong.** `verse_key` is `UNIQUE`, so the duplicate `INSERT` *rejects* (SQLITE_CONSTRAINT); it does not write a second row. With no try/catch, the rejection skipped both the optimistic store update and `SheetManager.hide`, which is the frozen sheet. Now `INSERT OR IGNORE` (bookmarking is idempotent by intent) plus a try/catch around the add/remove loop, with `SheetManager.hide` deliberately left *outside* the try so no persistence failure can strand the sheet.

**4 — fair, and I don't have a strong claim here.** `BOOKMARK_HIGHLIGHT_COLOR = HIGHLIGHT_COLORS.yellow` does make a bookmarked-only verse indistinguishable from an explicitly-yellow one, which does blunt the "give bookmarks a trace" intent. I've left the token alone in this PR since it's a design call rather than a correctness one — happy to add a distinct token here if you have a value in mind, or leave it for a follow-up.

**Nits:** duplicate `pageNumber` dep dropped; `loadedSurah` removed (grep confirms no consumers outside the store and its test); subset test reworked as above. One thing beyond the literal review: the `inFlightLoad` comment still ended with "leaving the store on the wrong key", which named a field this commit deletes, so I reworded that tail.

`tsc` 0 · jest 144/144 (8 in the reworked store suite) · eslint 0 errors, and one fewer warning than before from the removed dep. No `prettier --write` run — `SkiaPage.tsx` carries a pre-existing drift hunk ~590 lines from my edit, so I left it rather than reflowing unrelated code into the diff.

omar-zarka:
On finding 4 (bookmark tint === yellow highlight) — pushed a fix in `b61354c` that points the bookmark at its own token, a teal `rgba(64, 190, 190, 0.34)`, so it no longer collides with the yellow highlight. But the exact color is a **proposal, not a strong claim** — you own the palette, so here's the reasoning and the alternatives in case you'd prefer a different one (any of these is a one-line swap of `BOOKMARK_HIGHLIGHT_COLOR`).

**Constraint:** the marker has to stay distinct from all five user-selectable highlights (`yellow/green/blue/orange/purple`), so it can't just be another entry from `HIGHLIGHT_COLORS` — that only moves the collision. It needs a hue outside that set, which leaves roughly teal, rose/pink, or a neutral.

**Options I compared (light `#f4f3ec` and dark `#050b10` page):**

| | value | notes |
|---|---|---|
| **A — teal** (shipped) | `rgba(64, 190, 190, 0.34)` | distinct from all 5; reads clearly on the light page |
| B — rose | `rgba(237, 147, 177, 0.32)` | the most unambiguously "not a highlight" |
| C — neutral | `rgba(158, 152, 138, 0.42)` | reads as "saved" rather than "colored", but easiest to miss |
| D — edge bar | no fill; `border-inline-start: 5px solid #1D9E75` | a *different treatment* rather than a recolor |

**One thing worth flagging:** the tint is a fixed `rgba` applied regardless of theme, so on the dark page every *fill* option (including today's yellow) goes quite faint — they're tuned for the cream page. Option **D** is the only one that stays legible on both, because a solid edge bar doesn't rely on a translucent wash. It's also arguably the more correct model — a bookmark is a *different kind of thing* than a highlight (a saved location vs. a chosen emphasis), so treating it differently rather than just recoloring removes the confusion more durably. It's a bit more Skia work than a token swap, so I didn't assume it — happy to do D instead if that's the direction you'd want.

Teal (A) is in as the minimal, in-keeping fix. Just say the word if you'd rather have rose, neutral, the edge-bar treatment, or a specific value from your own palette.

PR #315, opened 2026-07-23
fix(player): mini-player a11y + in-flight-load race guard

omar-zarka:
Two independent fixes to the now-playing mini-player surface (`FloatingPlayer` on Android, `MiniPlayer` on iOS) and the store that backs it.

Ported from a downstream fork that carries these identical inherited components and hit both on real devices.

> **Scope change:** this PR originally also added a close (✕) control. Per review that's dropped — the mini player stays non-dismissable by design — along with everything that existed only to serve it (`playerStore.stop()`, its `usePlayerActions` binding, `LockScreenService.clearMainPlayer()`, and the `'stopped'` arm in `ExpoAudioProvider`'s merge guard). Those three files are now byte-identical to `develop`. The spacing tweaks that rode along with the button are reverted too — they were motivated by the ✕ crowding the play control, and a visual change nobody asked for doesn't belong here. What's left is the two pieces you said were worth keeping.

## 1. Screen-reader controls

The row was a grouping `Pressable` (`accessible=true`), so TalkBack/VoiceOver collapsed the whole bar into a single node — play/pause was only reachable by expanding to the full player.

The row is now a plain `View`, with the expand action moved onto an inner labeled `body` Pressable and play/pause left as an individually focusable, labeled sibling. **No visual delta:** `body` carries the same `gap` as the row it was split out of, and no other style value changes.

## 2. In-flight load race

A load is network-bound, so anything that lands while `updateQueue()` awaits `loadTrack()` can be overtaken by it: the stale load's `play()` fires afterwards and its caller's trailing `set()` clobbers the newer state. Two real cases on `develop` today:

- **Overlapping play taps / rapid re-queue** — two loads in flight, the older one wins the tail and you get a double `play()` against the wrong track.
- **`cleanup()` during a load** — the load plays into, and writes state over, a service that was just torn down.

A monotonic `playOpId` serializes them: `loadTrackAtIndex` claims an id on entry and re-checks it after each `await`, returning `false` when superseded; `cleanup()` bumps it synchronously; all five call sites skip their trailing state write when superseded. Net effect is last-legitimate-operation-wins, with a single `play()`.

Covered by a 3-test regression suite, `playerStore.loadRace.test.ts` (renamed from `stopRace` now that `stop()` is gone): the `cleanup()` race, a newer-load-supersedes-older race, and a normal load that must still finalize. I checked the suite has teeth — neutering the guard fails the cleanup test.

## Tooling note (for your own a11y verification)

While verifying #1 on-device: the mini-player subtree is **invisible to `uiautomator dump`** — the dump reflects `getRootInActiveWindow()`, which returns only the focused screen's window root, so an overlay/accessory painted by a separate RN surface never appears in the snapshot. Maestro and the live TalkBack/VoiceOver cursor both see and focus the controls fine. A snapshot-tooling artifact, **not** an a11y blackout — device-proven on a physical Pixel 3. Worth knowing if you ever script a11y checks off `uiautomator`.

## Verification

- `tsc --noEmit` — clean
- `jest` — full suite green, 34/34 (the one failing suite, `utils/__tests__/storageAnalytics.test.ts` "must contain at least one test", is pre-existing on `develop` and untouched here)
- `prettier --check` — clean on every touched file
- **Device-verified** on a physical Pixel 3 (release) in the fork carrying these identical inherited components: play/pause is individually focusable under TalkBack, and normal play/skip/queue flows are unaffected. The iOS-native `BottomAccessory` presentation was verified on a TestFlight build.

The two mini-player components are byte-identical to `develop` here, so the fork's device run is a faithful proxy for the UI change; the store change is exercised by the regression suite.

### DIFF
--- a/components/player/v2/FloatingPlayer/index.tsx
+++ b/components/player/v2/FloatingPlayer/index.tsx
@@ -1,6 +1,7 @@
 import React, {useCallback, useMemo, useRef} from 'react';
 import {View, Text, Pressable, StyleSheet} from 'react-native';
 import {usePathname} from 'expo-router';
+import {Feather} from '@expo/vector-icons';
 import {useTheme} from '@/hooks/useTheme';
 import {moderateScale} from 'react-native-size-matters';
 import {usePlayerActions} from '@/hooks/usePlayerActions';
@@ -23,7 +24,7 @@ import {useResponsive} from '@/hooks/useResponsive';
 export const FloatingPlayer: React.FC = React.memo(function FloatingPlayer() {
   const {theme} = useTheme();
   const glassColorScheme = useGlassColorScheme();
-  const {play, pause} = usePlayerActions();
+  const {play, pause, stop} = usePlayerActions();
   const playbackState = usePlayerStore(state => state.playback.state);
   const queueTracks = usePlayerStore(state => state.queue.tracks);
   const currentIndex = usePlayerStore(state => state.queue.currentIndex);
@@ -64,6 +65,10 @@ export const FloatingPlayer: React.FC = React.memo(function FloatingPlayer() {
     }
   }, [playbackState, pause, play]);
 
+  const handleDismiss = useCallback(() => {
+    stop();
+  }, [stop]);
+
   const containerStyle = useMemo(
     () => ({
       position: 'absolute' as const,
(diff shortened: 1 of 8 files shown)


osmansaeday (changes requested):
Thanks for this, and for the a11y and race-guard work. On the mini-player close button, though, we're going to pass, so please drop the `X` control from both `MiniPlayer` and `FloatingPlayer`.

Reasoning: the floating player doesn't occlude any content, so a dismiss affordance isn't solving a real problem, and it adds a way to lose the current playback/queue state for little gain. The platform-standard pattern in comparable apps (Spotify, Apple Music) is a persistent, non-dismissable mini player, and we want to match that. So the player stays unclosable by design.

The other two pieces are worth keeping and would be easy to land on their own:
- the accessibility improvements on the player controls, and
- the in-flight-load race guard in `playerStore` (+ its `stopRace` test).

If you can split those out of the close-button change (or just strip the close button from this PR and keep the rest), we'll move them forward. Marking as changes-requested only for the close button removal.

omar-zarka:
Dropped the close button — you're right, and we've taken the same decision downstream: the fork is removing its ✕ too rather than carrying a divergence here. A persistent mini player is the correct call.

Pushed in `053dd59` + `12eaaf3`. The ✕ is gone from both components along with everything that only existed to serve it:

- `playerStore.stop()` and its `PlayerStoreState` declaration (the button was its only caller)
- the `stop:` binding in `usePlayerActions`
- `LockScreenService.clearMainPlayer()`
- the `'stopped'` arm in `ExpoAudioProvider`'s merge guard

`ExpoAudioProvider.tsx`, `LockScreenService.ts` and `usePlayerActions.ts` are now **byte-identical to `develop`** — their only content in this PR was close-button machinery. `'stopped'` itself already existed as a `StorePlaybackState` member, so I reverted only the guard, not the type.

**I also reverted the spacing tweaks**, which you didn't ask about but which don't survive the reasoning: the padding/gap changes were motivated by the ✕ crowding the play control ("cramped against the right edge" was *with* the button beside it). With it gone that rationale goes too, and a visual change nobody asked for has no business in a PR that's now a11y + a race guard. Back to `develop`'s values — and the new `body` row carries the same `gap` it was split out of, so the a11y restructure has **zero visual delta**. Net: 7 files → 4.

What's left is the two pieces you wanted:

**a11y** — the row is a plain `View`, expand moved to an inner labeled `body` Pressable, play/pause an individually focusable labeled sibling. I checked removing the ✕ didn't quietly collapse the row back into a grouping `Pressable` (with one child that refactor would look redundant, but it's still load-bearing — a wrapping `accessible` Pressable would swallow play/pause back into a single node).

**Race guard** — kept whole. Worth noting it never depended on the close button: `cleanup()` is now its explicit invalidation source, and the "newer load supersedes an older in-flight one" claim inside `loadTrackAtIndex` is a standalone fix for overlapping play taps / rapid re-queue. Every comment that named `stop()` or the ✕ has been rewritten around those two.

**Test** — `git mv`'d to `playerStore.loadRace.test.ts` and the `describe` retitled, since `stopRace` names code that no longer exists. The `stop()` case is replaced by an equivalent **`cleanup()`** race, which matters more than a rename: `cleanup()` is now the only direct cover for `invalidatePlayOps()`. I verified that deliberately — neutering `invalidatePlayOps()` fails the new cleanup test, while the newer-load test still passes (it exercises the `++playOpId` claim inside `loadTrackAtIndex`, not the invalidation). So without that swap the guard's invalidation path would have gone uncovered. The other two cases are unchanged.

`tsc` 0 · `jest` 34/34 (the `storageAnalytics` "must contain at least one test" failure is pre-existing on `develop`) · `prettier --check` clean on every touched file. Title and description updated to match the new scope.

